PR #237, opened 2026-06-18
fix(linux): fix GTK type inference and MPV render runtime bugs

AdityaHebballe:
Hey! I know the official Linux build is still in progress but I was trying to get Harbor running on Arch Linux in the meantime and packaged it for the AUR ([harbor-stremio-git](https://aur.archlinux.org/packages/harbor-stremio-git)). While getting it working I ran into a couple of issues and fixed them, figured I'd send them upstream in case they're useful.

What this fixes:

- GTK raw pointer type inference — the compiler couldn't infer the type of two to_glib_none() calls; added explicit Stash<> annotations to fix compilation on some Rust toolchain versions.
- Black video on some setups — added epoxy_get_proc_address as a fallback GL proc loader (tried before eglGetProcAddress/glXGetProcAddress), and call attach_buffers() before the FBO query so the GL context has a valid framebuffer bound.
- Memory leak + crash when toggling MPV — the old teardown called mem::forget on the stale embed, which leaked memory and could crash on re-open. Replaced with a proper restore_webview() that fully reverses the widget reparenting.
- Video position/size wrong under fractional scaling — resize_to() now takes the CSS viewport dimensions and derives the correct CSS→native scaling factors from the actual GTK window allocation, so the video overlay lines up correctly.

### DIFF
--- a/src-tauri/src/mpv.rs
+++ b/src-tauri/src/mpv.rs
@@ -677,9 +677,11 @@ pub async fn mpv_set_geometry(
             let y = geom.css_top;
             let w = geom.css_width;
             let h = geom.css_height;
+            let css_view_w = geom.css_view_w;
+            let css_view_h = geom.css_view_h;
             let (tx, rx) = std::sync::mpsc::sync_channel::<()>(1);
             let _ = app.run_on_main_thread(move || {
-                let _ = crate::mpv_render_linux::resize_to(x, y, w, h);
+                let _ = crate::mpv_render_linux::resize_to(x, y, w, h, css_view_w, css_view_h);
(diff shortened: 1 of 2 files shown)


harborstremio:
@AdityaHebballe Thank you! We will revisit this to see what has already been done but this is super helpful thank you so much!

harborstremio:
Hey we really appreciate your effort and for packaging it on our behalf, we will be merging this for you as it is all good to us! Will not affect what we are working on and we will use your implementations instead of ours so you have that contribution for your efforts! Thank you again @AdityaHebballe

Issue #352, opened 2026-06-22
My Library doesn't work


ItsLincov:
my library doesn't work, once I click on "My Library" it goes directly to:

<img width="1081" height="535" alt="Image" src="https://github.com/user-attachments/assets/7885be5e-a5ee-415e-b470-8a783831c4d5" />

I have sent the report btw

nyouness:
Yes i can confirm this bug too

bqv6:
i fix it

harborstremio:
@nyouness  send me the log in technical detail pls

harborstremio:
oh i see it happened to us when we went really fast okay we will fix this

Code: Error
Title: Crash
Message: Minified React error #185; visit https://react.dev/errors/185 for the full message or use the non-minified dev environment for full errors and additional helpful warnings.

Detail:
Error: Minified React error #185; visit https://react.dev/errors/185 for the full message or use the non-minified dev environment for full errors and additional helpful warnings.

Error: Minified React error #185; visit https://react.dev/errors/185 for the full message or use the non-minified dev environment for full errors and additional helpful warnings.
    at Es (http://tauri.localhost/assets/index-DNzuh0t9.js:9:27671)
    at Va (http://tauri.localhost/assets/index-DNzuh0t9.js:9:27196)
    at zi (http://tauri.localhost/assets/index-DNzuh0t9.js:9:60108)
    at wu (http://tauri.localhost/assets/index-DNzuh0t9.js:9:59732)
    at ref (http://tauri.localhost/assets/index-DNzuh0t9.js:34:57066)
    at Ki (http://tauri.localhost/assets/index-DNzuh0t9.js:9:94411)
    at td (http://tauri.localhost/assets/index-DNzuh0t9.js:9:98793)
    at Hr (http://tauri.localhost/assets/index-DNzuh0t9.js:9:107118)
    at td (http://tauri.localhost/assets/index-DNzuh0t9.js:9:98757)
    at Hr (http://tauri.localhost/assets/index-DNzuh0t9.js:9:107118)

Component stack:

    at img (<anonymous>)
    at div (<anonymous>)
    at Poster (http://tauri.localhost/assets/index-DNzuh0t9.js:34:56177)
    at div (<anonymous>)
    at div (<anonymous>)
    at et (http://tauri.localhost/assets/library-BbwN-5Uj.js:2:8943)
    at div (<anonymous>)
[...]

PR #354, opened 2026-06-22
fix: poster blur-up + cached/failed handling; no render loop (#352)

bqv6:
@harborstremio  hey sorry for what happend, here the fix


harborstremio:
no worries at all! it happens we shoulda looked at it a lil closer andd thought of mac. our fault too :D we will close this as we put in a fix

Issue #318, opened 2026-06-21
1080p video does not scale to fill viewport on 4K displays


kfko:
making new issue as #179 was closed even though same issue persists.

When playing a 1080p video on a 4K monitor, the video is not scaled to fill the available viewport. The video is rendered only in the top-left portion of the player window (see attached screenshot).

- Monitor resolution: 3840×2160 (4K)
- Harbor version 0.9.26

I'm running auto player engine, though when checking if the issue occurs on both mpv and HTML5, I could not even play the video on HTML5, it would just go back to stream selection. I guess that's expected since it's .mkv file.

<img width="3840" height="2160" alt="Image" src="https://github.com/user-attachments/assets/2875a5e2-05c0-456a-8756-565fe341d740" />

https://github.com/user-attachments/assets/31b7a22d-b065-49bf-9002-5fb577d0f9ff

harborstremio:
@kfko Hey sorry for the issue again and for the inconvenience onthe ticket! We needed to clear things up. We will visit this for you and will put something in the upcoming betas please test it and let us know, to confirm are you on windows?

kfko:
@harborstremio I'm on macos. I'll be happy to confirm if any fixes you put out work, just let me know

harborstremio:
> [@harborstremio](https://github.com/harborstremio) I'm on macos. I'll be happy to confirm if any fixes you put out work, just let me know

Got it! that helps a ton we will push some patches out for you

harborstremio:
@kfko Hey! can yougo in settings > report a bug> export player log and send the harbor-mpv-log.txt file that gets placed in your downloads? Will help us fix this issue for you without regressing others let me know :) thank you

kfko:
@harborstremio sure, attached the logs
[harbor-mpv-log.txt](https://github.com/user-attachments/files/29239531/harbor-mpv-log.txt)

PR #399, opened 2026-06-24
fix: mpv macos viewport

kfko:
fix for: https://github.com/harborstremio/harbor/issues/318

I didn't see it breaking anything else, but if it does, at least you get the idea where the issue is.

https://github.com/user-attachments/assets/30293374-2a47-432f-b05b-8d7edc483bc2

### DIFF
--- a/src-tauri/src/mpv.rs
+++ b/src-tauri/src/mpv.rs
@@ -715,8 +715,8 @@ pub async fn mpv_set_geometry(
     {
         let x = geom.css_left;
         let y = geom.css_top;
-        let w = geom.css_width;
-        let h = geom.css_height;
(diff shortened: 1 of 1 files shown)


kfko:
I've looked a bit more into this and main culprit seems to be [this zoom](https://github.com/harborstremio/harbor/blob/3b0e35a75468dc232ba1f0885a8b866392bd68be/src/index.css#L3)

```css
@media (min-width: 3000px) {
  #root {
    zoom: 1.25;
  }
}
```

harborstremio:
Hey @kfko sorry for the delays! Thank you for taking this one on yourself to fix this really helps tremendously we will put in the fix manually as-well but will merge so u get contri

PR #477, opened 2026-06-27
fix(ui): resolve stuttering and lag during row navigation

### LINKED ISSUE #473: UI stuttering when navigating between content rows
I’ve been experiencing UI stuttering since the first day I started using Harbor.

The issue only affects the interface when navigating between content rows/lists (scrolling horizontally or moving between sections). Video playback is completely smooth, so the problem seems to be limited to the UI rendering.

Expected behavior

Navigation between rows and lists should be smooth without noticeable stuttering.

Actual behavior

The interface stutters while browsing the content, even though playback performance is perfectly fine.

Additional information

* The issue has been present since my first installation of Harbor.
* It is not related to a recent update.
* Video playback is smooth; only the UI is affected.
* GPU: RTX 2060 Super.
* Windows 11 Pro 22H2.
* I have already tried restarting the application, but the issue persists.

If you need logs or any additional information, I’d be happy to provide them

qpanx:
Closes #473

## Description
This PR addresses the UI stuttering and frame drops experienced when navigating or scrolling horizontally between content rows/lists, as reported in #473.

## Changes Made
- Optimized the rendering logic/components responsible for handling the content rows to prevent blocking the main thread.
- Reduced CPU overhead during rapid navigation, resulting in a smooth 60fps scrolling experience.

## Verification & Testing
Tested locally on **Linux (Fedora, WebKitGTK)** using DevTools Timelines. 
- **Before:** Noticeable UI lag and stuttering during horizontal list scrolling.
- **After:** Stable CPU usage graph, zero dropped frames, and buttery smooth rendering under rapid navigation.

### DIFF
--- a/src-tauri/Cargo.lock
+++ b/src-tauri/Cargo.lock
@@ -2017,7 +2017,7 @@ dependencies = [
 
 [[package]]
 name = "harbor"
-version = "0.9.18"
+version = "0.9.36"
 dependencies = [
  "axum",
  "base64 0.22.1",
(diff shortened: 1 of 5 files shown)


harborstremio:
Hey thanks for the help!

PR #532, opened 2026-06-30
Feat: Add configurable shared seek step

mazenS1:
## Summary

Adds a shared configurable seek step for keyboard arrows and player seek buttons using Harbor's existing settings and player control paths.

- Back and forward seek lengths are now stored in Harbor settings
- Keyboard `ArrowLeft` and `ArrowRight` use the configured Back and Forward seek steps
- Default player seek buttons use the same shared Back and Forward seek steps
- Custom player chrome seek icons also use the shared Back and Forward seek steps
- The Hotkeys settings Behavior section now exposes compact Back and Forward seek step controls
- Existing hidden `localStorage` seek step values are migrated when valid
- The fixed `10s` hotkey labels were replaced with neutral seek labels so the UI does not imply a hard coded duration

## Notes

This uses the existing settings store, hotkey system, and player transport control paths. No new dependency, native API, or platform specific input handling was added.

The shared options are `5s`, `10s`, `15s`, `30s`, `60s`, and `90s`. The default remains `10s` for both Back and Forward.

Legacy values from `harbor.seek step.back` and `harbor.seek step.forward` are read during settings load if they match the supported options. Invalid or missing values fall back to `10s`.

The transport button picker and the new Hotkeys settings controls now write to the same settings fields. Changing the value in either place updates the behavior for keyboard arrows, default seek buttons, and custom seek icons.

The separate `30s` shortcuts remain unchanged. Those bindings still seek by exactly `30s`.

## Discussion

This change intentionally keeps seek duration separate from key rebinding. The Hotkeys list still controls which keys trigger seek back and seek forward, while the new Behavior control decides how far those actions jump.

I also kept PiP and Kids transport seek buttons unchanged because they currently expose their own fixed seek behavior and were not part of the existing shared seek step path. That avoids expanding this PR beyond the player controls that already had configurable seek step behavior.

image

image

image


harborstremio:
great work! we will add this shortly

PR #550, opened 2026-07-01
Feat: Add player volume feedback indicator

mazenS1:
## Summary

Adds a transient volume indicator while watching, so hidden player controls still give feedback when volume changes.

## Changes

- Shows a compact volume HUD only when player chrome is hidden.
- Handles keyboard volume up/down, mouse wheel volume, mute, and player volume controls.
- **Volume up/down now unmute automatically.**
- Uses linear volume fill for HTML5 playback.
- Uses Harbor’s booste volume scale/colors for mpv playback above 100%.
- Keeps HTML5 hidden volume changes capped to 100%, matching visible controls.

image

image

image


harborstremio:
@mazenS1 hey this is great thanks for this sorry for the delays we will merge and integrate it here

PR #560, opened 2026-07-01
ci: add Windows ARM64 (aarch64) build to tauri-build

kipavy:
## What

Adds a **Windows ARM64** (`aarch64-pc-windows-msvc`) leg to the `tauri-build`
matrix, built natively on GitHub's `windows-11-arm` runner. This expands the
existing Windows x64 coverage to Windows-on-ARM without changing how any other
platform is built or released.

## Design

Every existing x86_64 / macOS / Linux step is **byte-for-byte unchanged** — all
new logic is gated to the `aarch64-pc-windows-msvc` target. The Rust side needed
no changes: `src-tauri/build.rs` already branches on `CARGO_CFG_TARGET_OS`, so it
was already correct for the ARM64 Windows target.

The ARM64-specific bits, mirroring what the x64 path does:

- **libmpv:** the committed `src-tauri/libmpv/mpv.lib` is x64-only, and Harbor
  hosts no bare ARM64 `libmpv-2.dll`. The ARM64 leg fetches a **pinned** upstream
  ARM64 libmpv dev bundle (shinchiro `mpv-winbuild-cmake`, tag `20260610`,
  checksum-verified) and regenerates the MSVC import library for arm64 from the
  already-committed `libmpv.def` via `lib /machine:arm64` — the same approach as
  `setup-libmpv.ps1`. `ilammy/msvc-dev-cmd` (arch `arm64`) provides `lib.exe`.
- **ffmpeg / ffprobe:** fetch BtbN's `winarm64` builds instead of `win64`.
- **yt-dlp:** keep the x64 `.exe` (no native ARM64 build; it runs under
  Windows-on-ARM emulation), named for the aarch64 triple so the sidecar resolves.
- **mpv.exe:** the auto-merged `tauri.windows.conf.json` bundles an `mpv.exe`
  resource, so the ARM64 leg provides the ARM64 player build.

## Validation

I cross-compiled this target locally (Linux, cargo-xwin) using the exact same
dependency steps and it **compiles, links, and runs**: a native ARM64
`harbor.exe` launched and worked on Windows-on-ARM (libmpv, WebView2, etc. all
functional; PE machine type `0xAA64` verified on every binary except the
intentionally-x64 `yt-dlp.exe`).

## Please confirm in CI

The `tauri-build` workflow is `workflow_dispatch`, so it doesn't run on PRs.
Everything except the CI-runner environment specifics is proven locally — could a
maintainer **dispatch the workflow on this branch** to confirm the new leg is
green? The only pieces not exercised locally are the `windows-11-arm` hosted
runner, `ilammy/msvc-dev-cmd` (arm64), and `lib.exe`/`7z` availability on that
runner. Happy to adjust based on the run.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/.github/workflows/tauri-build.yml
+++ b/.github/workflows/tauri-build.yml
@@ -15,6 +15,10 @@ jobs:
             runner: windows-latest
             target: x86_64-pc-windows-msvc
             bundle: msi
+          - name: Windows-ARM64
+            runner: windows-11-arm
+            target: aarch64-pc-windows-msvc
+            bundle: msi
           - name: macOS-AppleSilicon
             runner: macos-latest
             target: aarch64-apple-darwin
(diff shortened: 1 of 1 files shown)


kipavy:
@harborstremio When will it be available for Windows ARM in releases ?? Thanks

Issue #522, opened 2026-06-30
Embedded mpv player renders in wrong size/position on fractional Wayland scale (niri, scale 1.25)


troyjvv:
i know not officially released on niri, found on aur. used claude to help with bug details and this writeup. am using new version of harbor.

Environment:

OS: Arch Linux (CachyOS)
Compositor: niri (Wayland)
Display: 1920x1200, output scale set to 1.25
Harbor: built from source / latest release 

Bug description:
When using embedded mpv playback mode with a fractional output scale (1.25), the video only renders in a small region in the top-left corner of the window instead of filling it. Switching to "separate window" mpv mode works fine at any scale. Switching the niri output scale to an integer value (1.0 or 2.0) also fixes embedded mode completely.
Root cause (from logs):
At scale 1.25, the mpv_linux backend reads the surface scale as integer 2 instead of 1.25, and allocates a render surface based on that wrong multiplier:
[harbor::mpv_linux] render surface 960x912 px scale 2
[harbor::mpv_linux] render surface 1536x960 px scale 2
At scale 1.0 (same screen, same video), the surface is calculated correctly:
[harbor::mpv_linux] render surface 16x16 px scale 1
[harbor::mpv_linux] render surface 1920x1200 px scale 1
1536x960 is exactly 1920x1200 divided by the real scale factor of 1.25, while the surface itself is logged as scale 2. This points to the Wayland backend rounding/flooring fractional scale to the nearest integer rather than reading the actual fractional value (likely via wp_fractional_scale_v1), then using that wrong integer to compute the render surface buffer size.
Steps to reproduce:

On a Wayland compositor (niri in this case), set output scale to a non-integer value, e.g. 1.25
Launch Harbor, enable embedded mpv mode (not separate window)
Play any video
Video renders only in a small region, top-left of the window, rest of the window is blank

Workaround:
Setting output scale to an integer (1.0 or 2.0) fixes embedded playback. Separate window mode also works fine regardless of scale.
Suggested fix:
Read fractional scale via wp_fractional_scale_v1 if supported by the compositor, instead of falling back to integer wl_output scale, when computing the mpv embed surface size on Linux/Wayland.

harborstremio:
Hey we are sorry for the delays! We are gonna put in a patch for you but linux is "technically" not officially released yet so we do apologize for the inconveniences and the issues :(!

PR #594, opened 2026-07-03
fix(mpv_linux): fix fractional Wayland scale causing video to render in top-left corner (using scale other than 1)

troyjvv:
Already made an issue that got closed. i know this isn't the place to do it fully since linux not fully supported, but i found the linux file and made a fix and this would make it easier to implement.

Fixes issue #522

 removed integer scale_factor() multiplication in do_render fallback and resize_to, which was rounding 1.25 (my scale)  to 2 and allocating an oversized render surface


harborstremio:
Hey we will merge it we appreciate your work anyways and will bundle current linux soon while people wait for full release

PR #635, opened 2026-07-05
feat: add IMDb API fallback for cast-modal when no TMDB key

SlasshyOverhere:
Add free imdbapi.dev API as a fallback metadata source for the 'About this title' modal when users have not configured a TMDB API key.

Changes:
- New provider: src/lib/providers/imdbapi/imdbapi-details.ts Fetches title details, cast, crew, and images from api.imdbapi.dev. Returns a TmdbDetail-compatible shape so it slots into the existing modal with zero UI changes. Caches results to avoid duplicate fetches.

- Settings: added 'imdbApiFallback' boolean toggle (default: true) in types.ts and defaults.ts.

- Settings UI: added ToggleRow in Library & metadata panel, right
  below the TMDB key field. Label: 'Use free IMDb API as fallback'.

- CastModal: added third branch in useEffect — when no TMDB key is set but toggle is ON and meta.id starts with 'tt', fetches from imdbapi.dev instead of showing the 'Add a TMDB key' prompt.

- CastStrip: fixed image src to handle full URLs from imdbapi.dev (checks startsWith('http') before prepending IMG/w185 prefix). Poster fallback already handled this — no change needed.

Behavior:
- TMDB key set → TMDB (unchanged)
- No TMDB key + toggle ON → imdbapi.dev (new)
- No TMDB key + toggle OFF → 'Add a TMDB key' message (unchanged)
- Anime titles → Kitsu/AniZip (unchanged)

Screenshot 2026-07-05 222320

### DIFF
--- a/src/components/player/cast-modal.tsx
+++ b/src/components/player/cast-modal.tsx
@@ -2,6 +2,7 @@ import { Loader2, Star, Users, X } from "lucide-react";
 import { useEffect, useState } from "react";
 import type { Meta } from "@/lib/cinemeta";
 import { animeDetails } from "@/lib/providers/anime-detail";
+import { imdbapiDetails } from "@/lib/providers/imdbapi/imdbapi-details";
 import { useSettings } from "@/lib/settings";
 import { useT } from "@/lib/i18n";
 import { activeLayout } from "@/lib/theme";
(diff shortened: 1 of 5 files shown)


harborstremio:
@SlasshyOverhere Hey! sorry for the delay and thanks for the PR! We will look at this and merge this shortly

PR #642, opened 2026-07-06
Add Android support for Harbor

Lawtro37:
This is a **VERY** in development port of Harbor to Android (specifically arm32a Android TV)

Currently compiles and shows up as an app when installed ~~but crashes instantly upon launch~~.
Actively being worked on by me but could definitely use some help if you wish to contribute.

Ideally this should be to be merged back into the original Harbor repo once its at a stable state.
For now I will leave this as a draft.

**UPDATE!**
NOW LAUNCHES ON ADROID TV!!!
sucsess
(image slightly edited for privacy)
The UI needs to be modified to fit on the screen correctly and work with a TV remote.


harborstremio:
This is awesome we will give it a peek and appreciate your contributions!

Lawtro37:
NOW LAUNCHES ON ADROID TV!!!

Lawtro37:
@harborstremio based on the mobile notice that gets displayed on mobile devices, "This instance of Harbor is made for desktop. Our standalone iOS and Android apps are coming soon, each with a bespoke, mobile-first experience built for its native platform." I should rebuild or at least heavily modify the web UI for this.
I have been trying to maintain build parity, as in all versions of Harbor can be built from the same source).
What I'm thinking I can do is create a separate source folder called `src-android` that the android build will use instead of changing what the default source displays based on the platform.
This would make future updates a pain as they would need to be implemented for both sources but would make the code cleaner and smaller.
I could possibly even split the sources of mobile android devices and TV android devices.
How would you recommend/want me to go about this?

harborstremio:
> @harborstremio based on the mobile notice that gets displayed on mobile devices, "This instance of Harbor is made for desktop. Our standalone iOS and Android apps are coming soon, each with a bespoke, mobile-first experience built for its native platform." I should rebuild or at least heavily modify the web UI for this. I have been trying to maintain build parity, as in all versions of Harbor can be built from the same source). What I'm thinking I can do is create a separate source folder called `src-android` that the android build will use instead of changing what the default source displays based on the platform. This would make future updates a pain as they would need to be implemented for both sources but would make the code cleaner and smaller. I could possibly even split the sources of mobile android devices and TV android devices. How would you recommend/want me to go about this?

Well we where gonna do per platform but we paused that a while ago to build the main desktop one as it had launch bugs we wanted to sort out. We do not mind doing it as a cross compile thing and like making a UI for androids though like thats a non issue and probably the easiest solution for others to maintain in the long run. So if this is something you genuinely are gonna work on right now we will be working on this build with you once we get main desktop version to stable. We do not want to undo any of your work and value your efforts seriously! We do not have the man power most days to keep things pushing on other platforms so this is exactly what harbor needs even if we got work to do

harborstremio:
So this is great seriously @Lawtro37 we also accepted the repo inv we will push some stuff there when we can

Lawtro37:
I cant figure out why but the app keeps using the desktop source files even though I know the android source files are getting built to `/dist`. 
I might need to take a sort break from this project (just for a couple of days or so) to work on some other important stuff.

harborstremio:
> I cant figure out why but the app keeps using the desktop source files even though I know the android source files are getting built to `/dist`. I might need to take a sort break from this project (just for a couple of days or so) to work on some other important stuff.

no problem you can take a break or stop any time as-well you are not obligated to work on it and we appreciate all your efforts :D we are gonna be doing our own mobile and desktop regardless if you chose to work on this we would take your baseline and run with it thank you again for the PR and the effort!!

O1Abdulrahman:
Have you used Capacitor? It’s easy to use, it worked fine with no crashes and problem. You just need some enhancements to make it work smoothly without issues when playing media, using some plugins.

I also developed a navigation system that works with TV devices, and it’s working well. You can use it in the app if you’d like.

Lawtro37:
> Have you used Capacitor? It’s easy to use, it worked fine with no crashes and problem. You just need some enhancements to make it work smoothly without issues when playing media, using some plugins.
> 
> I also developed a navigation system that works with TV devices, and it’s working well. You can use it in the app if you’d like.
> 

I will look into using Capacitor, looks promising.
Also yes a working TV navigation system would be amazing as I am not very familiar with react. If you want, could you make a fork and PR of my repo to add the TV UI to the scr-android directory and the new packages to the package.json. that feels like the easiest most natural way to do it.
Thank you for the suggestion and developing a working TV UI.

PR #697, opened 2026-07-09
i18n: Add Brazilian Portuguese (pt-BR) translation

heltonfonias:
Hi there! 

I have translated the interface strings into Brazilian Portuguese (`pt-BR`) to help localize the application. 

I also added a File inside of locales folder called pt.ts, but I don't know if it went through. 

Please let me know if any adjustments are needed. Thanks!

PR #698, opened 2026-07-09
Added Portuguese Brazil localization file for i18n

heltonfonias:


harborstremio:
#697 We will merge both of these

PR #714, opened 2026-07-10
keyboard / remote TV navigation

O1Abdulrahman:
Added TV-style keyboard/remote navigation to make directional focus movement more reliable across the app, including arrow-key navigation and WASD, back-button handling, and better focus behavior for sidebar/content transitions.
space key to Enter the film , ESC to back.
it's worked in TV remote as well.


harborstremio:
thank you for the contribution and good feature!

PR #722, opened 2026-07-10
Fix keyboard and TV remote navigation

aajsa:
## Summary

Fix and integrate the new keyboard/TV remote navigation hook so Harbor builds correctly and directional navigation works without overriding existing keyboard controls.

## Background

`src/App.tsx` imported `./useKeyboardNavigation`, but the hook was added at the repository root. Because the file was outside `src/`, Vite could not resolve the import and the application failed to start.

The hook also registered a global capture-phase keyboard listener. That listener received Escape and arrow keys before search, dialogs, the player, and keyboard-aware widgets could handle them. This caused several regressions beyond the original missing-file build error.

## Changes

- move `useKeyboardNavigation.ts` into `src/` so the existing import resolves
- keep spatial arrow/WASD navigation enabled for normal page content
- leave inputs, textareas, selects, and editable elements in control of their own keys
- allow Escape to reach the search overlay and close it after search is opened with `/`
- disable global spatial navigation while the player is active so arrow keys continue to seek and change volume
- defer keyboard handling to widgets that implement their own navigation, including listboxes, menus, grids, trees, and tab lists
- detect the active dialog and keep spatial focus within it instead of moving focus to controls behind the overlay
- allow dialogs to handle Escape before the global page-back behavior
- identify the search overlay as an accessible modal boundary

## Expected behavior

- `/` opens global search and Escape closes it
- arrow keys seek and control volume normally in the player
- menus and Live TV category lists retain their own arrow-key behavior
- Escape closes the active dialog instead of navigating the page behind it
- directional keyboard/remote navigation continues to move between visible controls on regular pages
- focus cannot move behind an active modal

## Verification

- keyboard ownership regression checks pass
- editable-control Escape regression check passes
- `pnpm build` passes

## Manual test checklist

- open search with `/`, type a query, then close it with Escape
- enter the player and verify left/right seeking and up/down volume
- open menus, dialogs, and the Live TV category list and verify their local keyboard controls
- press Escape inside a dialog and confirm the dialog closes without changing the underlying route
- navigate regular pages using arrow keys or a TV remote and confirm focus remains visible and directional


O1Abdulrahman:
Thanks bro, I really appreciate this! 🙌 😊

harborstremio:
Great work thank you :D

PR #723, opened 2026-07-10
Add phone remote control UI over web server (`/remote`)

jaynav1:
## Summary

I wanted a way to control Harbor playback from my phone. Thanks to #714 I was able to add navigation as well. This adds the `/remote` route as a host navigation and playback remote. 

The remote shows an Apple TV-style touchpad for scrolling through Library items and a back button. Tapping the touchpad selects the item in focus. Still working on ironing out navigation bugs.
When playing media, the poster and details are shown. Controls for play/pause, volume, and seeking are exposed.

Useful for:

1. **HTPC connected to TV**: browse and play from the couch without a keyboard and mouse.
2. **Headless / cast setups**: longer-term path toward a full phone UI.

Let me know whether this is worth continuing or if you have other plans for browser features.

## Changes

- WebSocket remote on the LAN web server (`/api/remote`) + phone UI at `/remote` (touchpad, now playing, transport, renderer picker)
- Settings copy clarifies `/` vs `/remote` under the same “Harbor in your browser” toggle; Vite proxy so `/remote` hot-reloads in `tauri`dev
- Full-screen text entry when the host is focused in a text field with keyboard passthrough

Also made some UI tweaks to facilitate navigation based on my limited testing:

- Moved the #714 hook into `src/lib` so it typechecks and App + remote share it
- Phone touchpad reuses controls (swipe = arrows, tap = click focused control, Back sends host Back)
- Better focus zones (sidebar / hero / content), modal scopes (search, local episode pickers), default focus after page changes, and a visible focus ring
- Back from player always exits playback (previously streaming multiple episodes all entered navigation stack)
- Search top match: only Open is focusable, not the whole card. Side effect is top match card is no longer selectable, might not be best solution

## Screenshots
| **Browsing touchpad** | **Text entry page** |
| ------------- | ------------- |
|Screenshot_20260711_015824_Chrome|Screenshot_20260711_015858_Chrome|

|**Now playing controls**|**Casting sheet**|
| ------------- | ------------- |
|Screenshot_20260711_015956_Chrome|Screenshot_20260711_020013_Chrome|

## Notes

Player hotkeys interfere with keyboard navigation, may look into separating them. Casting from the remote (casting sheet → cast device) hasn’t been tested either.

## Test plan

- Phone opens `/remote` on LAN; renderer sheet + transport work
- Touchpad: swipe / tap / back; search traps focus; keyboard on host still works
- Text entry: Done / Go behave; top match Open focuses and activates
- Next episode then Back once → leaves player; remote transitions back to browsing touchpad
- Background the phone app and return after a while; test if connection is re-established
- Navigate around the UI using the touchpad and see if it feels natural, some pages may need more changes.


harborstremio:
This is super cool! we will merge as soon as you are ready

jaynav1:
Added player and picker navigation by creating new focus events and using those instead of arrow keys which trigger hotkeys. Also ported and added nav changes after incorporating #740 and #743. Mainly adding classes to modals so they can be navigated/exited, and some more tweaks. Will test a bit more before deeming ready.

agscs:
this is supper cool

O1Abdulrahman:
This is great. I will be with you. If you face any problem, I will fix it.

PR #737, opened 2026-07-11
fix: make Tauri listener cleanup idempotent

aajsa:
## Summary
- Ensure Tauri event listener unlisten is safe to call more than once.
- Apply in pause-on-inactive player hook to avoid double-unlisten races.

## Test plan
- [ ] Open/close player and background the app repeatedly
- [ ] No console errors from unlisten / dispose


aajsa:
Closing to test locally before reopening.

PR #738, opened 2026-07-11
fix: consolidate keyboard navigation hook

aajsa:
## Summary
- Consolidate keyboard/TV remote navigation into `src/lib/use-keyboard-navigation.ts`.
- Remove the leftover root `src/useKeyboardNavigation.ts` duplicate.
- Support `enabled` and ownership boundaries so dialogs/search can own Escape.

## Test plan
- [ ] Arrow-key focus moves across home rails
- [ ] Escape closes search / dialogs without fighting global back
- [ ] Navigation disabled while player is open


aajsa:
Closing to test locally before reopening.

PR #741, opened 2026-07-11
fix(bug-report): reliable diagnostics and file attachments

aajsa:
## Summary
- Start error capture at app boot so reports include recent runtime errors
- Detect OS via UA/platform helpers (removed broken `@tauri-apps/plugin-os` import)
- Count installed addons in diagnostics instead of hardcoding `0`
- Clearer network / player-log export errors
- Fix file preview object-URL leak; i18n for drop zone reject messages

## Test plan
- [ ] Settings → Bug report: submit a short report successfully
- [ ] Attach image/video under limits; reject oversized/non-media with readable messages
- [ ] Export player log after playing media; before playing shows friendly empty-log message
- [ ] Diagnostics card shows OS + addon count


aajsa:
Closing — not ready to publish yet (local only).

PR #780, opened 2026-07-12
fix(networking): contain DNS load to stop overwhelming systemd-resolved

JJDizz1L:
## Summary
Harbor was overwhelming the host system's DNS resolver, causing system-wide network hangs. With Tailscale's MagicDNS in the path, the combination of one-`reqwest::Client`-per-HTTP-call, a 100-QPS DHT, and torrenting defaults being on was enough to push MagicDNS into "degraded feature set" mode. Manually restarting `systemd-resolved` and `NetworkManager` was the only recovery. Investigation of 10 h of journal logs showed 44k+ DNSSEC validation failures and 410 unique reverse-DNS lookups originating from Harbor's HTTP traffic, the torrent engine, and Steam P2P combined.

This PR adds connection reuse, DNS caching, and (most importantly) makes Harbor's HTTP traffic bypass `systemd-resolved` entirely via reqwest's `hickory-dns` feature.

## Changes
- `8ed5661` — `harbor_fetch` now uses a shared `OnceLock<reqwest::Client>` instead of one per call. Connection pooling, no more fresh DNS lookups for every HTTP call.
- `884062c` — DHT query rate reduced from 100 QPS to 20.
- `c583ad8` — `netcheck` uses a DNS cache and a shared HTTP client.
- `1dcd430` — Replaced `expect()` panics with proper `Result`/`Option` returns in the shared-client initializers.
- **`9e58cea` — The key fix:** enable reqwest's `hickory-dns` feature on both the `http_fetch` and `stream_proxy` clients. This bypasses `systemd-resolved` for Harbor's HTTP traffic entirely. hickory-resolver talks directly to upstream DNS and has its own local cache. Net effect: Harbor's HTTP DNS load no longer competes with the rest of the system for MagicDNS bandwidth.

## Files changed
- `src-tauri/Cargo.toml` — added `hickory-dns` to reqwest features
- `src-tauri/src/http_fetch.rs` — shared client + `hickory_dns(true)`
- `src-tauri/src/stream_proxy.rs` — `hickory_dns(true)`
- `src-tauri/src/torrent_engine.rs` — DHT rate 100 → 20
- `src-tauri/src/torrent_engine/netcheck.rs` — DNS cache + shared client

5 files changed, 78 insertions, 19 deletions.

- **Tested on Linux** (Arch-based, system-resolved + Tailscale MagicDNS).
  - Built with `npx tauri build --no-bundle` — note: plain `cargo build --release` does **not** embed the frontend and will produce a blank window. Always use the Tauri build command.
  - DNS load on `systemd-resolved` observed to drop to ~2.6 queries/sec with ~68% cache hit rate, vs the prior baseline of 700+ queries/min.
- **NOT yet tested on Windows / macOS.** Suggested verification:
  - Windows: DNS resolution path through hickory-resolver (Windows uses a different DNS stack than systemd-resolved).
  - macOS: same items as Windows, plus `torrentsDisabled` lazy re-read against macOS file-locking on `settings.json`.
  - Suggested: build, launch, run for an hour while browsing Stremio catalogs, and confirm system DNS load matches the Linux baseline.

## Compatibility
- No new external Rust dependencies beyond `hickory-resolver`, which is a transitive dep of reqwest already.
- No frontend dependency changes.
- Backward compatible: no default behavior change for end users.
- Cross-platform code: every change is in shared `src-tauri/src/...` (no `cfg(target_os = ...)` blocks needed).

## Follow-up
A second PR (`fix/torrents-killswitch`, builds on top of this one) adds a user-facing `torrentsDisabled` kill-switch. It is split into a separate PR for faster review of the DNS fix.


JJDizz1L:
This image show the FLOOD of torrent queries that causes network instability (at least on Linux) causing confusion to users as their Internet no longer functions. This is a necessary improvement to get this application to function properly on a Linux system.
image

This image is showing the result of torrents disabled via user kill-switch and the DNS query list significantly shortened and has been stable for over 10 hours of testing on a Linux system.
image

harborstremio:
Thank you for the help! :D

PR #781, opened 2026-07-12
feat(torrents): add 'Disable torrents entirely' user kill-switch

JJDizz1L:
## Summary
Adds a new global `torrentsDisabled` setting that prevents the torrent engine from ever starting. When enabled, the engine session is never created, so DHT, side DHT, trackers, UPnP, and the stream server all stay off.

This builds on top of #780 (DNS containment). On its own, this PR is a user-facing UX improvement: it gives users a hard kill-switch independent of the existing `directTorrentStream` setting (which only gates the picker, not the engine).

## Changes

**New setting** `torrentsDisabled: boolean` (default `false`) added to the `Settings` type and the default settings blob.

**Rust side** — `src-tauri/src/settings_store.rs`:
- New `read_torrents_disabled(app) -> bool` helper. Cheap field scan on `settings.json` (no full JSON parser — just looks for `"torrentsDisabled":true`). Returns `false` if the file is missing or the field is absent.

**Rust side** — `src-tauri/src/torrent_engine.rs`:
- `ensure_started_on_setup` checks `read_torrents_disabled()` and early-returns with a log message + sets `last_error`. No engine init at all.
- `ensure_session` (the lazy-init path for mid-session `torrent_engine_add` calls) returns an error if disabled.
- Lazy re-read: every call to these functions re-checks the flag, so toggling the setting takes effect without an app restart.

**Frontend** — `src/lib/torrent/stremio-stream.ts`:
- New `torrentsDisabled()` helper that reads from `localStorage`.
- `directTorrentEnabled()`, `directStreamAvailable()`, `engineP2pEligible()` all return `false` when `torrentsDisabled()` is true. Frontend defense-in-depth layer.

**UI** — `src/views/settings/player-panel/p2p-advanced-section.tsx`:
- New "Disable torrents entirely" toggle with explanatory sub-text.
- Amber warning text under the toggle when enabled.
- `directTorrentStream` toggle is now locked with reason *"Disabled because torrents are disabled above"* when the new flag is on.
- Toggling ON fires `torrent_engine_hard_reset` to stop the engine immediately.

**UI** — `src/views/play-picker.tsx`:
- New amber banner at the top of the picker with an "Open Settings" link to direct the user to flip the toggle back on.

## Files changed
- `src-tauri/src/settings_store.rs` — `read_torrents_disabled` helper
- `src-tauri/src/torrent_engine.rs` — `torrentsDisabled` gates on `ensure_started_on_setup` and `ensure_session`
- `src/lib/settings/defaults.ts` — `torrentsDisabled: false` default
- `src/lib/settings/types.ts` — `torrentsDisabled: boolean` field
- `src/lib/torrent/stremio-stream.ts` — `torrentsDisabled()` helper + gating
- `src/views/play-picker.tsx` — amber banner
- `src/views/settings/player-panel/p2p-advanced-section.tsx` — new toggle

7 files changed, 114 insertions, 1 deletion.

- **Tested on Linux:**
  - Toggling `torrentsDisabled = true` confirmed to stop all torrent engine activity (DHT quiet, no tracker traffic, no listening ports, no stream server).
  - Toggling back to `false` re-enables the engine on the next call.
  - Picker shows the amber banner with the "Open Settings" link, and the `directTorrentStream` toggle is correctly locked.
- **NOT yet tested on Windows / macOS.** Suggested verification:
  - Windows: `torrentsDisabled` gate behavior, "Open Settings" banner in the picker.
  - macOS: same items, plus the `torrentsDisabled` lazy re-read in `ensure_session` against the macOS file-locking behavior of `settings.json`.

## Compatibility
- New global setting with safe default (`false` = current behavior). Existing users are unaffected.
- `directTorrentStream` setting is unchanged and untouched.
- Cross-platform code: no `cfg(target_os = ...)` blocks needed.
- No new dependencies.

## Depends on
This branch is based on `fix/dns-stability` (#780). It can be rebased to main if the DNS PR is merged first, or merged as-is to preserve the full change history.

Screenshots of the new toggle, the amber warning, and the picker banner will be added shortly.


JJDizz1L:
Again, the purpose of the PR is for user whom prefer use of their debrid services over the use of torrenting directly through peer-to-peer. This ensures that no torrent connections occur when a debrid API exists in the user's settings and will show warnings about torrents being disabled in multiple locations.

image
image

harborstremio:
great work :D!

PR #798, opened 2026-07-13
Keyboard Navigation Improvements, And SFX

O1Abdulrahman:
### Keyboard Nav

-Add a color indicator to the search bar in Settings to show whether you're in Text Input mode or Navigation mode.
ESC to Exist text mode, Enter to enter text mode. and add SFX when you inter and Exist.

-IF YOU mouse click anywhere you will Exist NAV mode.

-Fixed the home search when inter the search in home whether via Key-Nav or mouse you will enter the text mode , ESC to Exist **AS USUAL**, NO nav mode.

and Fix ESC in Keyboard when you enter the film via mouse click, you can go back via ESC.

### Audio Feedback

🔊 Added a new theme -Cloudy-
-fixed old theme problem,

⚙️ Settings & UI Enhancements

-Smart Volume Slider: Added a dedicated SFX volume slider in the Settings panel.
-Real-time Feedback: The slider plays a short audio tick when dragged, allowing users to preview the volume level instantly.

🎬 Video Player Micro-Interactions

-Volume Hotkeys Feedback: Added extremely short audio ticks when adjusting the video volume using hotkeys or the slider. Each theme features a custom pattern.

-Pause Audio Cue: Added a distinct sound effect that plays when pausing the video, Each theme features a custom pattern

-mute as well

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -469,13 +469,16 @@ function Shell() {
     onBack: handleTvBack,
     onBackToNav: handleTvBackToNav,
   });
-useEffect(() => {
+  useEffect(() => {
     if (settings.soundTheme) {
       SFX.setTheme(settings.soundTheme);
     }
-  }, [settings.soundTheme]);
-  useEffect(() => startMaintenance(), []);
-
+    
+    const volume = settings.sfxVolume ?? 50; 
+    
+    SFX.setVolume(volume / 100); 
+    
+  }, [settings.soundTheme, settings.sfxVolume]);
   useEffect(() => {
     const initAudio = () => SFX.init();
     window.addEventListener("pointerdown", initAudio, { once: true });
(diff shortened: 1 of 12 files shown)


jaynav1:
Hey there, love the sound work so far and they work great with #723. `keyboard-navigation.ts` and `use-keyboard-navigation.ts` both export a full `useKeyboardNavigation` and seem to have drifted, which one do you intend to keep? Main currently imports `@/lib/keyboard-navigation`, which I assumed for my work.

I understand the Text input mode is your solution for inputting WASD as text. I never encountered this conflict because I have been using arrow keys, and since moved to separate focus handlers for my remote to work with the player UI. The modes may introduce friction for arrow-key / TV / remote users.

My suggestion is removing WASD as directional input for now, and extracting the solution for WASD input into a follow-up. Happy to help with directional input bugs from my testing as well.

O1Abdulrahman:
thanks 😊, for now the keyboard-navigation file, and i add a request for new update, that fixed the wasd problem. and I'll take your suggestion into consideration.
It's also a great feature that you're working on.

aajsa (changes requested):
#

aajsa (approved):
#

aajsa (changes requested):
Thanks again for working on this. Would it be better to avoid checking English and Arabic text in `aria-label` values and use language-independent markers instead? This will keep Back navigation and SFX working correctly if we change or add UI languages. Also, seek actions should not use the volume-change sound.

These are my review notes. Thank you again.

O1Abdulrahman:
> Thanks again for working on this. Would it be better to avoid checking English and Arabic text in `aria-label` values and use language-independent markers instead? This will keep Back navigation and SFX working correctly if we change or add UI languages. Also, seek actions should not use the volume-change sound.
> 
> These are my review notes. Thank you again.

thanks I almost forgot it. i will try to fix it

harborstremio:
Hey @OakArbum We are gonna implement this for ya manually on beta tree (while theres conflicts here) you can resolve conflicts and we will merge or close up to you this is great though :D!

O1Abdulrahman (commented):
grate

PR #814, opened 2026-07-14
fix: set_bundled_mpv commented out as it was removed in thumbs.rs

Ayman-Hachkal:
thumbs.rs set_bundled_mpv function was gutted for clean-up, which caused an error when building due to lib.rs calling that function.

### DIFF
--- a/package.json
+++ b/package.json
@@ -11,6 +11,7 @@
     "preview": "vite preview",
     "tauri": "tauri",
     "tauri:build:linux-system": "tauri build --config src-tauri/tauri.linux-system.conf.json",
+    "setup": "node scripts/fetch-binaries.mjs",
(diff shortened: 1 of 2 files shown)


AdityaHebballe:
Thanks for the fixes. I made two small follow-up cleanups on this PR:

* Removed the stale Windows set_bundled_mpv setup block. Tauri’s Windows resource directory is the executable directory, and locate_mpv() already searches for the bundled mpv.exe there.
* Renamed the new script to setup:binaries so it doesn’t collide with PNPM’s built-in pnpm setup command. Use pnpm run setup:binaries to fetch the required build sidecars.

The binary bootstrap script is useful for fresh clones because those bundled Tauri sidecars are gitignored. Looks good now. I'll merge this.

PR #848, opened 2026-07-15
fix(kids): stop restricted content persisting when switching to a kid…

hzeemr:
… profile

Switching into a kid profile while a page was already open (an adult title's detail page, its related titles, and its stream sources) left that content fully visible and playable: the kid-mode view guard allowed meta/grid/collection views but nothing reset the navigation stack on entry, so a pre-open page survived the switch.

- Reset to the Kids home when a kid profile becomes active; setView("kids") clears the whole stack, back history included, so no pre-open page persists.
- Filter the kids detail "More to explore" rail with the same genre exclusion the kids catalogs use, and populate genres on TMDB recommendation/similar items so the filter is effective instead of a no-op.

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -774,7 +774,14 @@ function Shell() {
 
   useEffect(() => {
     if (!activeProfile) return;
-    if (!activeProfile.kid && topKind === "kids") setView("home");
+    if (activeProfile.kid) {
+      // Reset to the Kids home so a page already open (e.g. an adult title and
+      // its related titles) cannot persist across the switch into a kid profile;
+      // setView("kids") clears the whole navigation stack, back history included.
+      setView("kids");
+    } else if (topKind === "kids") {
+      setView("home");
+    }
   }, [activeProfile?.id]);
 
   const playerActive = !!player;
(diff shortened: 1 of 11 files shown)


aajsa:
Thanks @hzeemr for tracking this down and fixing the navigation-stack leak when switching into a kid profile. I tested the original scenario and confirmed your reset-to-Kids approach works well.

I added a few follow-up commits directly to this PR branch:

- preserved TMDB's `adult` flag through catalog, recommendation/similar, and collection mappings
- kept your original Horror/Thriller filtering for “More to explore”, while separately excluding `adult: true` items across Kids heroes, rows, recommendations, franchise grids, and collection grids
- kept Cinemeta available when no TMDB key is configured, with a conservative kids-only Family/Animation filter
- ensured full collection navigation stays in Kids mode and cannot reveal filtered collection items

Verification completed with TypeScript, project lint/type checks (no errors), an independent code review, and a live Cinemeta check confirming that `1776` and Demon Slayer are excluded while Moana and Toy Story remain.

Thank you again for the solid foundation and for contributing this fix. I’ll merge it with your original commit and authorship preserved.

PR #893, opened 2026-07-17
Fix rtx hdr, add keybind and player button

Justsnoopy30:
Sorry for the format churn, that came from vp format which is weird because I would think that would have already been run by now. The original diff was only 200.

Also makes sure to respect scoped SVP settings & restores changed settings when toggled off.


AdityaHebballe:
RTX HDR hotkey/button doesn’t apply the required D3D11 decoder configuration to an already-running player. The new pre-init path selects hwdec=d3d11va only when rtxHdr is true at bridge creation https://github.com/harborstremio/harbor/blob/2d9d2f4779f744ceb82356c5d01ba3e99c27b451/src-tauri/src/mpv.rs#L289

But the bridge key excludes playerRtxHdr, so toggling the new UI control or Ctrl+H does not recreate mpv; it only attempts to add the video filter 

https://github.com/harborstremio/harbor/blob/2d9d2f4779f744ceb82356c5d01ba3e99c27b451/src/views/player/hooks/use-player-bridge.ts#L74

A player launched with RTX HDR off therefore remains on hwdec=auto, despite the UI reporting RTX HDR active. Since the implementation explicitly requires d3d11va for RTX HDR, the advertised in-playback toggle is unreliable and can silently do nothing. Include this setting in the bridge/restart decision, or avoid presenting it as a live toggle unless the active decoder is compatible.

Please clarify the intended behavior as the UX can be misleading

Justsnoopy30:
I don't think it actually matters as mpv defaults to d3d11va on windows anyways when using hwdec=auto. It was just to be more explicit in case it ever changed but it's perfectly functional.

AdityaHebballe:
If you have tested it and it works then it is fine. Good work!

PR #898, opened 2026-07-17
fix(native): harden resource lifecycle

aajsa:
## Summary
- make cast HLS sessions own and terminate their ffmpeg processes, with bounded sliding playlists
- bound thumbnail caches and native IPC queues, move blocking cleanup work off async hot paths, and add cancellable torrent cache sweeping
- add bounded mpv event-error backoff plus visible player failure handling
- make Tauri listener cleanup idempotent so stale listener callbacks do not become crash reports
- tighten macOS renderer teardown without changing the cross-platform playback architecture

## Verification
- `vp check <changed files>` (0 errors; existing React warnings remain)
- `rustfmt --check` on all changed Rust files
- `pnpm test` (36 passed)
- `vp run typecheck`
- `pnpm build`
- `cargo test --manifest-path src-tauri/Cargo.toml` (32 passed)
- `cargo test --manifest-path harbor-core/Cargo.toml` (56 passed)

## Notes
- Native OpenGL deprecation warnings on macOS are pre-existing; replacing the renderer with Metal is intentionally out of scope.
- Platform-specific process/AppKit behavior is compile- and unit-tested here; hardware integration still benefits from Windows/Linux smoke testing.


AdityaHebballe:
When a new cast starts, clean up the previous HLS session after the new device accepts its load. ACTIVE is overwritten with the new session, while stop_hls_session is only called on explicit cast_stop or failed startup. Switching casts can therefore leave the prior ffmpeg process and its temp directory alive until the five-minute idle GC. The new HLS ownership work is good, this closes its remaining handoff gap.
Could fix this in the PR but we'd also need to test two HLS-backed casts consecutively and assert the first session is stopped.

AdityaHebballe:
Pushed in b905bb3 (fix(thumbs): stop shadow workers on app shutdown).

Adds reliable thumbnail-mpv shutdown to the lifecycle cleanup path.

Thumbnail workers are now stopped for both main-window exit and tray → Quit. Cleanup sends quit with try_send, escalates to killing only the shadow mpv process, and bounds the post-kill wait so shutdown cannot hang on a stalled IPC writer.

The change also serializes in-flight thumbnail worker creation with shutdown, preventing a just-spawned worker from escaping before it is registered in state.

macOS/Linux verification:
- Before: pgrep -af 'mpv.*harbor-thumbs-' showed 4 orphaned workers after Harbor exit.
- After: the same check returned no workers after both window-close and tray-Quit paths.

Normal playback mpv is not affected.

aajsa:
> Pushed in [b905bb3](https://github.com/harborstremio/harbor/commit/b905bb3aea9e2ff7f728c1fbb5114f1dd3b653a1) (fix(thumbs): stop shadow workers on app shutdown).
> 
> Adds reliable thumbnail-mpv shutdown to the lifecycle cleanup path.
> 
> Thumbnail workers are now stopped for both main-window exit and tray → Quit. Cleanup sends quit with try_send, escalates to killing only the shadow mpv process, and bounds the post-kill wait so shutdown cannot hang on a stalled IPC writer.
> 
> The change also serializes in-flight thumbnail worker creation with shutdown, preventing a just-spawned worker from escaping before it is registered in state.
> 
> macOS/Linux verification:
> 
> * Before: pgrep -af 'mpv.*harbor-thumbs-' showed 4 orphaned workers after Harbor exit.
> * After: the same check returned no workers after both window-close and tray-Quit paths.
> 
> Normal playback mpv is not affected.

Pulled b905bb3 and verified the thumbnail shutdown changes on the current PR head.

I also confirmed the HLS handoff gap: after a new device accepts its load, ACTIVE replaces the previous session without stopping the previous HLS session. I’ll fix this by atomically replacing ACTIVE, releasing the mutex, then stopping the previous HLS session, with a regression test covering two consecutive HLS-backed casts.

While validating that test path, I found a pre-existing ffprobe parsing issue: the current output emits codec_name before codec_type, while the parser expects codec_type first. On current ffprobe this leaves video_codec empty and causes HLS registration to fall back with “no video stream found”. I’ll fix the parser so the consecutive-cast test exercises the real HLS path.

I also found that the new Child-based HLS ownership has no shutdown/drop cleanup for active temp directories. I’ll add an explicit HLS shutdown path so active ffmpeg sessions and harbor-hls-* directories are cleaned when Harbor exits.

Does this approach look right to you, or would you prefer the pre-existing ffprobe parser fix to be handled in a separate PR?

AdityaHebballe:
I think the same PR would be good. Good job optimising harbor!

aajsa:
Implemented the follow-up in this PR.

- Fixed ffprobe parsing so stream codecs are associated correctly whether `codec_name` or `codec_type` is emitted first. Both field orders have regression coverage.
- After a new device accepts its load, Harbor now atomically replaces `ACTIVE`, releases the mutex, then stops the previous HLS session. The consecutive-cast test also verifies cleanup runs without holding `ACTIVE`.
- Added idempotent application-shutdown cleanup that drains every HLS session, terminates and reaps its ffmpeg child, removes its `harbor-hls-*` directory, and only then marks the exit clean.

Verification:
- `cargo test --manifest-path src-tauri/Cargo.toml`: 43 passed
- `pnpm test`: 46 passed
- `vp run typecheck`
- `pnpm build`
- PR CI: Frontend and Native passed

Issue #901, opened 2026-07-17
Tone down the buffering indicator in player chrome


AdityaHebballe:
<img width="720" height="134" alt="Image" src="https://github.com/user-attachments/assets/434d36e3-9735-4fc2-a678-0034d7f8be7e" />

@aajsa

The buffering indicator introduced in #899 renders the animated multicolor Harbor loader directly over the center play button. It is visually much louder than the surrounding player chrome and makes the transport look broken while buffering.

Could we tone this down? A subtle buffering/progress ring around the play button would retain the feedback without obscuring the control or drawing so much attention.

Reproduction: start playback, wait for a confirmed buffering stall, then reveal the player controls.

aajsa:
@AdityaHebballe 

Thanks for catching this, you’re right. The full Harbor loader was too visually loud and overlapped the center control. I’ve opened #902 to replace it with a subtle neutral loading ring that keeps the play/pause icon visible, while preserving the existing 300ms confirmed-stall behavior. Could you verify the updated design?

AdityaHebballe:
Yes let me build it, I'll let you know. Thanks for the fix!

harborstremio:
Hello gentlemen we will swoop this with our tree and implement your PRS here  when we do push beta sync if anything needs to be fixed or reimplemented let us know

PR #902, opened 2026-07-17
fix(player): tone down buffering indicator

aajsa:
## Summary

- replace the animated multicolor Harbor loader with a subtle neutral progress ring
- keep the center of the control clear so the play/pause icon remains visible
- preserve the existing 300 ms confirmed-stall behavior
- respect reduced-motion preferences

## Verification

- `pnpm test` (41 passed)
- `vp run typecheck`
- `vp check src/views/player/buffering-indicator.tsx tests/player-loading-state.test.ts`
- `git diff --check`

Refs #901

The issue should remain open pending reporter verification.

### DIFF
--- a/src/components/player/shells/minimal-shell.tsx
+++ b/src/components/player/shells/minimal-shell.tsx
@@ -36,7 +36,8 @@ export function MinimalShell({
         </button>
         <button
           onClick={onPlayPause}
-          className="flex h-12 w-12 items-center justify-center rounded-full bg-white/15 text-white backdrop-blur-md transition-colors hover:bg-white/25"
+          data-player-play-pause
+          className="relative flex h-12 w-12 items-center justify-center rounded-full bg-white/15 text-white backdrop-blur-md transition-colors hover:bg-white/25"
           aria-label={playing ? t("Pause") : t("Play")}
         >
           {playing ? (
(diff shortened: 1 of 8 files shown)


AdityaHebballe:
image
I still see the same animation for some reason. Can you share a screenshot of how it looks?
Maybe my build is cached

aajsa:
I don’t think it was a cache issue. I initially misunderstood the report because the animation wasn’t appearing on my setup. I’ve updated it now so the buffering ring is rendered directly around the actual play/pause button instead of floating over the center of the player. Could you please try the latest version and let me know if it looks correct now?

AdityaHebballe:
https://github.com/user-attachments/assets/7ac48763-94ee-44f8-8c89-13a830f54fbd

It's strange. This animation does not load correctly on my mac but yet appears on my linux. Your changes did not change the animation. On my mac it briefly appears but glitches out.

Is it supposed to be rainbow?

aajsa:
The rainbow effect in the video came from the existing LiquidGlassSurface used by the play button, not from the buffering indicator. The buffering indicator itself is only a thin white rotating ring. I’ve now set the play button’s spectralStrength to 0 in commit 4123e89, so the rainbow effect should no longer appear. Could you please try the latest build?

AdityaHebballe:
https://github.com/user-attachments/assets/4b0c4f94-4486-4419-95be-5c118119dade

It does look much better now but it still looks off to me, might just be me nitpicking honestly. 
image
The rainbow effect is still present in the home screen play button though.

Is it a rendering issue or the effect itself is supposed to look like that?

aajsa:
The rainbow effect is intentional and comes from the existing ThreeLiquidGlassSurface, but its current appearance and inconsistent rendering across platforms are not ideal. This is separate from the buffering indicator addressed by this PR. I’ll update the ThreeLiquidGlassSurface implementation soon, which should fully resolve it across the player and home screen.
Thank you for testing it and pointing this out.

PR #910, opened 2026-07-17
fix(thumbs): stop shadow workers on app shutdown

aajsa:
## Summary

- stop thumbnail shadow workers during application shutdown
- complete the remaining reliability lifecycle change after #898

The branch contains one commit not yet present in `main`.


aajsa:
Closing because the shadow-worker shutdown implementation is already present in main; the apparent difference came from the branch's older merged history.

PR #916, opened 2026-07-18
perf(ui): lazily activate thumbnail glass

AdityaHebballe:
## Summary

- activate thumbnail backdrop blur only after a card is first hovered or focused
- prewarm the first blur layer invisibly for one compositor frame, avoiding the initial frosted-surface pop
- retain instant reveals after the compositor is warm and keep carousel arrows unchanged

## Windows validation

@aajsa, could you please check the Home thumbnail play/remove controls on a Windows device? In particular, verify the first hover has no delayed blur, icons stay sharp, and there are no black-screen regressions.


AdityaHebballe:
@aajsa I have tested this on mac and linux and the effect looks visually good and efficient as well. #909 had made the effect disappear and hence it was visually difficult to distinguish the button from the background.

Please verify this and let me know if you face any issues.

PR #917, opened 2026-07-18
Liquid Glass enhancements, improvements, and more features.

O1Abdulrahman:
it is A hybrid 2D interface with a Three.js glass shader.


AdityaHebballe:
Hi your changes overlap with mine, I have optimised the liquid glass effect for performance as well as visuals. I will resolve the conflicts in you PR. Thank you

AdityaHebballe:
@OakArbum we should be careful changing the liquid glass implementation moving forward. Noticed a lot of performance regressions in this PR.

AdityaHebballe:
Closing this for now: it replaces the existing CSS liquid-glass surface with a Three/WebGL renderer, which regresses the behavior we need to preserve. In particular, the blur can become visible after the interaction begins and the new compositing path can blur foreground controls/icons. It also adds a canvas/rendering path and Three.js bundle cost without a clear visual improvement over the existing renderer.\n\nThe current renderer already gives us the required performance, immediate blur behavior, and crisp foreground controls by keeping the blurred surface behind an isolated content layer. If we add new glass controls or poster treatment later, please build them on that existing renderer/layering model rather than replacing it, and preserve those no-regression requirements.

Please build your changes upon the latest main and test for performance regressions.

AdityaHebballe:
Please build on the existing liquid-glass renderer rather than replacing it.

Requirements:

- Keep it CSS-based no Three.js, WebGL canvas, render loop, or poster texture processing.
- Preserve the current layer order: blur/glass surface behind an isolated foreground content layer. Foreground buttons, icons, text, and focus rings must never be included in the blur.
- Blur must be ready when the interaction becomes visible; don’t introduce a delayed “blur catches up” effect.
- Do not add per-pointer-move rendering, canvas resizing, image redraws, global texture caches, or persistent animation loops.
- New visual controls must have a visible, intentional effect and belong under the existing Liquid Glass settings section.
- Poster treatment should be a lightweight CSS frame/overlay, opt-in, and must not redraw or filter the poster image itself.
- Preserve current behavior and performance across normal pointer, keyboard, and remote/focus navigation.
- Keep the change narrowly scoped, avoid unrelated formatting churn or renderer rewrites.

Before opening a PR, verify that controls remain crisp, blur appears immediately on first interaction, and there is no measurable performance or bundle-size regression.

PR #919, opened 2026-07-18
liquid glass improvement and new Feature

O1Abdulrahman:
css with no Three.js.

### Keyboard navigation can now access the side navigation bar in every theme.


ricardosousa191:
The Three.js/canvas removal looks right, checked ThreeLiquidGlassSurface.tsx and confirmed no canvas/WebGL/requestAnimationFrame render loop remains. 

But this diff is 856 files, 32k lines, way beyond the liquid glass scope. Looks like formatting churn got mixed in with the feature branch. Could you rebase/squash against latest main so this is actually reviewable? 
As is, nobody can realistically verify there's nothing else hiding in the noise.

Also, most commits here are just "update", "U" with no description, which makes it impossible to review incrementally or bisect later if something regresses. 

Worth adopting descriptive commit messages going forward, even short ones.

AdityaHebballe:
The diff is huge for a PR that improves liquid glass. Vibe coding gone wrong moment.

AdityaHebballe:
Thanks for reworking this without Three.js/WebGL that addresses one of the major concerns with the previous approach.

I'm closing this revision because it still regresses the existing Liquid Glass behavior, and the scope is too broad to merge safely.

This PR changes 894 files and includes unrelated routing, query/virtualization, native Rust/WebView/mpv, build, and documentation changes. Please rebase onto the current `main` branch and submit a focused Liquid Glass PR only.

More importantly, the current renderer and Continue Watching implementation already solve several behavior and performance issues that we don't want to regress:

* Blur is ready immediately when the controls appear.
* Foreground icons, text, focus rings, and buttons remain crisp.
* Surfaces do not gain backdrop blur unless they explicitly opt into it.
* There is no visual or performance regression.

This revision removes the Continue Watching warm-up/activation behavior and adds a 220 ms opacity transition around the controls, which reintroduces the delayed blur/glass appearance that the current implementation already avoids. It also changes the renderer defaults to `variant="surface"` and `backdropBlur={true}`, causing backdrop blur to appear at call sites that previously did not opt into it.

Please build incrementally on the existing renderer rather than replacing its behavior:

* Preserve the existing isolated background glass layer and foreground content layer.
* Keep backdrop blur opt-in only, preserving the current defaults.
* Do not change the Continue Watching activation/warm-up path.
* Keep controls, icons, text, and focus rings above the blurred layer, and verify they remain sharp.
* Add only settings with a clear visual effect, along with sensible defaults and valid ranges.
* Include targeted verification for immediate first-hover blur, crisp foreground rendering, keyboard/focus navigation, and no measurable performance regression.

A small, rebased, focused PR that preserves the existing behavior while extending it would be welcome.

As @ricardosousa191 said please use descriptive commit messages and avoid making changes beyond the scope of the PR. Catching regressions would be harder and the quality of the codebase would go down over time.

image

PR #920, opened 2026-07-18
feat: adopt TanStack Query/Router and harden Windows WebView2 recovery

aajsa:
## Summary

- adopt TanStack Query, Router, and Virtual across catalog, discover, movies, and shows routes with intent preloading and progressive catalog loads
- recover from Windows WebView2 render-process death and stuck black screens, with fail-open startup and embedded mpv hwnd cleanup
- remove the memory-hud diagnostic overlay and its Ctrl+Shift+M shortcut
- pause the outgoing stream during source swaps so audio cannot leak behind the swap loader, and crossfade the player route fallback instead of a black blink
- sample per-process RSS on macOS for memory-pressure relief
- CI: move macOS-Intel builds to macos-15-intel (macos-13 runner retired)

## Root cause

The Windows App build broke because tauri 2.11 pulls webview2-com 0.38 while harbor pinned 0.37, so the ProcessFailed handler and the WebView2 controller came from different windows-core versions (E0277). A moved `Vec` in the mpv hwnd cleanup also failed the borrow check (E0382).

## Verification

- `cargo check --manifest-path src-tauri/Cargo.toml` (host)
- webview2-com 0.38 API verified for `x86_64-pc-windows-msvc` with a mirror crate covering every harbor call site; the full Windows-target check is blocked locally by aws-lc-sys cross-compilation, so CI covers it
- `vp run typecheck`
- `vp check` on changed files (repo-wide formatting drift is pre-existing and left untouched)

Windows runtime validation remains appropriate because the reported failure is platform-specific.


aajsa:
Superseded by focused splits:

- #921 — feat: adopt TanStack Query, Router, and Virtual
- #922 — fix(windows): prevent stuck WebView2 black screen
- #923 — fix(windows): reload WebView2 on render-process death
- #924 — refactor: remove memory-hud, pause outgoing stream on swap

PR #922, opened 2026-07-18
fix(windows): prevent stuck WebView2 black screen

aajsa:
## Summary

- keep WebView2 opaque during normal browsing and enable transparency only while embedded mpv is active
- fail open at startup if `PageLoadEvent::Finished` never arrives
- resume WebView2 after `TrySuspend` / `SetIsVisible(false)` so a suspended controller cannot remain invisible
- hide and destroy embedded-mpv child HWNDs after playback stops
- preserve ownership of the HWND list during cleanup

## Root cause

Always-on transparency over a black native HWND presents as a stuck black window when WebView2 composition fails. A suspended or invisible controller, or an orphaned embedded-mpv child window, can keep covering the interface until the entire app restarts.

## Behavior

- browsing uses an opaque WebView2 background
- transparency is limited to active embedded-mpv playback
- startup eventually reveals the main window even if page-load completion stalls
- hidden/suspended WebView2 controllers resume when returning to the UI
- stale embedded-player child windows are removed on stop

## Verification

- `cargo check --manifest-path src-tauri/Cargo.toml` on the host
- GitHub Frontend and Native checks passed before the latest main update and are being rerun
- Windows runtime validation remains required because the reported failure is platform-specific

## Review focus

- WebView2 visibility and transparency transitions
- startup timeout behavior on slow and failed loads
- HWND ownership and cleanup after embedded playback
- repeated play/stop and suspend/resume cycles


aajsa:
Superseded by #923, which now combines this PR's WebView2 visibility, transparency, startup fail-open, and embedded-mpv HWND cleanup with the native renderer-failure watchdog in one reviewable PR.

PR #924, opened 2026-07-18
refactor: stabilize stream switching and player transitions

aajsa:
## Summary

- pause the outgoing stream during source swaps so audio cannot leak behind the swap loader
- resume the existing stream when resolution or proxy registration fails before takeover
- clear the swap loader only from the latest active request
- crossfade the player route fallback instead of showing a black blink
- retain the memory HUD and its `Ctrl+Shift+M` shortcut
- sample per-process RSS on macOS for memory-pressure relief
- move macOS Intel CI builds to `macos-15-intel`

## Why

Overlapping stream switches could let stale requests control loader state, failed takeovers could leave playback paused, and route transitions could briefly flash black. The memory HUD also needs accurate per-process attribution on macOS to remain useful during playback diagnostics.

## Behavior and scope

- the current stream remains recoverable until a replacement successfully takes over
- only the latest swap request owns the transition loader
- player-route transitions use a visual fallback instead of a black frame
- memory diagnostics remain available and gain macOS process RSS sampling
- the macOS Intel workflow uses GitHub's current Intel runner label

## Verification

- `vp check` on changed frontend files
- `vp run typecheck`
- GitHub Frontend and Native checks pass
- platform-specific runtime validation remains appropriate for stream switching and macOS memory attribution

## Review focus

- overlapping or rapidly cancelled stream switches
- resume behavior after resolver/proxy failure
- latest-request ownership of loader cleanup
- macOS process attribution and memory-pressure thresholds

## Platform notes

Runtime validation is still requested on macOS for RSS attribution and on at least one native-player platform for source-swap timing.


AdityaHebballe:
- Stale swaps can still win. After registerStreamProxy() resolves, the code never re-checks ac.signal.aborted. If the user selects B while A is waiting on proxy registration, A can finish later and call b.load()/close the switcher, overriding the newer choice. Add an abort/generation check immediately after proxy registration and before every state-changing step.

- Retrying then failing leaves playback paused. The first selection pauses playback. A second selection captures the now-paused snapshot, so if that second resolve/proxy step fails, resumeOnFailure is a no-op. The previously playing stream remains paused. Track whether playback was active before the whole switch session, not per attempt, and resume it on a latest-attempt failure.

I fixed these 2 issues. We can merge this.

AdityaHebballe (approved):

PR #933, opened 2026-07-18
fix

aajsa:
## What changed

- make the Watch Together popover owned by only the button instance that opened it
- keep the button size and position stable while the popover is open
- keep page inputs and controls interactive while using Watch Together

## Root cause

Multiple mounted Watch Together buttons shared one global open flag. Every instance rendered a popover and registered an outside-click handler, so one instance could close another while the user interacted with its inputs. The open state also resized and translated the button wrapper.

## Validation

- `vp check` on the changed files
- `vp run typecheck`
- `pnpm test` (68 passed)
- production frontend build passed before publishing


aajsa:
Superseded by #934, which contains only the Watch Together fix on top of main.

PR #938, opened 2026-07-18
glass effect improvement and new Feature

O1Abdulrahman:
css with no Three.js.

### Keyboard navigation can now access the side navigation bar in every theme.


Justsnoopy30:
im convinced you must be trolling

PR #939, opened 2026-07-19
fix: cancel native fetches and media processes

aajsa:
## Summary

- propagate `AbortSignal` cancellation into native HTTP requests and cap buffered responses at 16 MiB without bypassing the cap through plugin fallback
- add managed timeout/cancellation cleanup for yt-dlp, FFmpeg, FFprobe, subtitle extraction, direct MPEG-TS streams, and cast subtitle preparation
- terminate Harbor-owned child processes and direct descendants on timeout across Unix and Windows paths
- serialize cast loads with generation ownership so stale requests cannot replace newer sessions, and clean stale device/HLS sessions
- stop stalled transcoding immediately when the HTTP response consumer disconnects

## Why

Long-running native requests and media subprocesses could outlive the operation or client that created them. This wastes memory and CPU, leaves orphaned work behind, and allows stale cast loads to replace newer sessions.

## Scope and behavior

- cancellation is propagated across the frontend/native fetch boundary
- response buffering is bounded at 16 MiB on every fetch path
- Harbor-owned subprocess trees are cleaned up on cancellation and timeout
- cast generations make the newest load authoritative
- disconnected transcode consumers stop the producer promptly
- normal successful playback, casting, subtitle extraction, and trailer flows remain unchanged

## Verification

- `vp check` on all changed files
- `vp run typecheck`
- `cargo test --manifest-path src-tauri/Cargo.toml --lib` — 52 passed
- `pnpm test` — 61 passed
- GitHub Frontend and Native checks pass
- independent code review found no remaining Critical or Important issues

## Review focus

- process-tree termination behavior on Unix and Windows
- cancellation ownership and cleanup on every early-return path
- cast generation ordering and stale-session cleanup
- enforcement of the 16 MiB response limit without fallback bypasses

## CI reliability fix

The intermittent `process::tests::timeout_terminates_direct_descendants` failure was a false positive: on Linux, a killed descendant can remain briefly as a zombie until PID 1 reaps it, and `kill(pid, 0)` reports that zombie PID as existing.

The assertion now inspects the process state instead. Zombie state `Z` or an absent PID counts as terminated; every genuine live state still fails the test.

Verification after the fix and latest `main` merge:

- targeted descendant-timeout test — 100/100 stress iterations passed
- `cargo test --manifest-path src-tauri/Cargo.toml --lib` — 52 passed
- `vp check src/lib/safe-fetch.ts`
- `vp run typecheck`
- independent review found no Critical, Important, or Minor issues


AdityaHebballe:
Unix cleanup kills only direct child processes, so a deeper helper process can survive and keep its pipes open.
I will verify if this is the case and commit a fix if true. Sorry for the delay!

AdityaHebballe (approved):

PR #940, opened 2026-07-19
glass effect improvement and new Feature

O1Abdulrahman:
Expanded Liquid Glass across the top bars, menus, search, and Watch Together.
css with no Three.js.

Added settings to enable or disable Liquid Glass and control its opacity.
Added optional Poster Dock magnification with configurable settings.
Keyboard navigation can now access the side navigation bar in every theme.


ricardosousa191:
Read through the diff, wanted to leave some notes.

The glass on/off toggle pattern is inconsistent across the PR. In back-chrome, control-renderer, search-overlay, continue-card and row.tsx, the whole element gets duplicated for the enabled/disabled case. But royal-topbar, topbar and together-modal do it the cleaner way, pull the content into a variable once and just swap the wrapper. Would be good to make that consistent everywhere.

Also noticed a leftover non-English comment in topbar.tsx ("الشكل الأصلي قبل إضافة Liquid Glass").

In ThreeLiquidGlassSurface.tsx there's spectral${"Strength"} as a weirdly split way to write "spectralStrength", could just be a plain string.

And no tests anywhere in this diff, magnified-dock-item.tsx especially has some non-trivial transform math that could use at least one. The popover positioning logic in royal-topbar, topbar and topdock is also duplicated three times almost identically, might be worth a shared hook down the line.

Just sharing what I noticed, not trying to block anything.

O1Abdulrahman:
> Read through the diff, wanted to leave some notes.
> 
> The glass on/off toggle pattern is inconsistent across the PR. In back-chrome, control-renderer, search-overlay, continue-card and row.tsx, the whole element gets duplicated for the enabled/disabled case. But royal-topbar, topbar and together-modal do it the cleaner way, pull the content into a variable once and just swap the wrapper. Would be good to make that consistent everywhere.
> 
> Also noticed a leftover non-English comment in topbar.tsx ("الشكل الأصلي قبل إضافة Liquid Glass").
> 
> In ThreeLiquidGlassSurface.tsx there's spectral${"Strength"} as a weirdly split way to write "spectralStrength", could just be a plain string.
> 
> And no tests anywhere in this diff, magnified-dock-item.tsx especially has some non-trivial transform math that could use at least one. The popover positioning logic in royal-topbar, topbar and topdock is also duplicated three times almost identically, might be worth a shared hook down the line.
> 
> Just sharing what I noticed, not trying to block anything.

Thank you so much for taking the time to review the diff and share these notes. I genuinely appreciate it.

Your feedback is very helpful, and I’ll work on addressing the inconsistencies and cleanup you mentioned. If you have access and are willing, please feel free to make changes directly on the PR branch and add any improvements you think would make the implementation cleaner or better.

AdityaHebballe:
It currently fails Frontend CI: 4 failing tests in tests/keyboard-focus.test.ts.

  - TV focus no longer waits for navigation intent.
  - Focus styling regressed from the required theme-aware accent ring.
  - Back handling is in the wrong order relative to the global keyboard eligibility guard.
  - A global search listener no longer uses the shared keyboard eligibility guard.

  That alone makes it unmergeable.

  The bigger issue is performance. This is not a lightweight CSS enhancement:

  - ThreeLiquidGlassSurface is now ~500 lines and renders five layered surfaces per glass instance, using backdrop filters, masks, filters, blend modes, gradients, and will-change.

  - When a surface is alwaysActive, which the persistent top dock and some menus are, it starts four infinite CSS animations: lens, spectrum, caustics, and sheen. That is a continuous compositor/GPU workload, even when the user is doing nothing.

  - The existing “no GPU render loop” test still passes because it only checks for canvas/WebGL/requestAnimationFrame; it does not detect these infinite CSS animations. So the test gives a false sense of safety.

  - It changes the surface defaults to variant="surface" and backdropBlur=true, so callers can now get expensive backdrop filtering without explicitly opting into it.
    
The dock magnification is unoptimized too:
  - Each pointer move updates a shared MotionValue.
  - Every rendered poster item responds by calling getBoundingClientRect() in its transform calculation.
  - On a row of many posters, that creates many layout reads per mouse event, plus springs/transforms per card. It needs measured/centralized geometry and throttling, not independent per-card DOM measurements in the pointer hot path.
  
  
This is highly unusable on any machine that doesn't have a decent GPU. The current implementation is lightweight CSS while maintaining visual clarity and the frosted glass effect. I feel performance should be a priority for a streaming app and visual effects must not come in the way of it. I will try to make the effect optional and opt-in instead and make changes to this PR itself, CSS can be default and your implementation can be an option. This will take time as even if it's optional we should ensure it shouldn't slow down the user's machine. Let's properly test it before merging

@ricardosousa191 @aajsa feel free to weigh in on this. Please check the current implementation of liquid glass on main and provide feedback. Thank you

AdityaHebballe:
Made some changes.

  - Keep the established Liquid Glass renderer as the default.
  - Add **Enhanced Liquid Glass** as an opt-in renderer with an opacity control.
  - Remove persistent glass animation loops from the enhanced renderer.
  - Add configurable default-glass blur (0–8px); 0 fully disables backdrop blur.
  - Make the search pill show its glass treatment consistently instead of dimming it until hover.
  - Add optional Poster Dock magnification with smoother row-level updates, no sideways movement as it adds jank, and enough top spacing to prevent poster clipping.
  - Remove unrelated formatting-only changes from the PR.
  
  Please test the effect now. Performance is good now

AdityaHebballe (approved):

PR #942, opened 2026-07-19
fix(search): prevent stale results and cache metadata

AdityaHebballe:
## Summary
- prevent stale search responses from replacing the active query and render only results that match it
- cache TMDB, Cinemeta, and anime metadata searches for 60 seconds with a bounded 16-entry cache per provider
- keep addon results uncached; show a partial-results notice when TMDB fails
- allow Linux debug builds to run alongside the installed Harbor app while release builds retain single-instance behavior

## Advantages
- avoids incorrect or missing Top Match/results after rapid query changes
- reduces repeated metadata API calls for backspacing, reopening, and refining searches
- bounds memory to 48 cached result sets and keeps variable addon results fresh
- preserves release single-instance behavior and adds a stale-response regression test

## Review notes
- reviewed the complete diff against ; no search behavior regressions found
- addon catalog searches are intentionally not cached
- the cache key scopes TMDB data by API key, language, genre filtering, and normalized query

### DIFF
--- a/src-tauri/src/lib.rs
+++ b/src-tauri/src/lib.rs
@@ -415,7 +415,11 @@ pub fn run() {
     let dvr_state = dvr::DvrState::new();
     let multiview_state = multiview::MultiviewState::new();
     let modal_overlay_state = modal_overlay::ModalOverlayState::new();
-    let app_builder = tauri::Builder::default()
+    let app_builder = tauri::Builder::default();
+    // Let a Linux development build run alongside the installed Harbor app.
+    // Packaged builds keep the normal single-instance behavior.
+    #[cfg(not(all(target_os = "linux", debug_assertions)))]
+    let app_builder = app_builder
         .plugin(tauri_plugin_single_instance::init(|app, args, _cwd| {
             use tauri::{Emitter, Manager};
             if let Some(w) = app.get_webview_window("main") {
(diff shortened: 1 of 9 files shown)


aajsa:
Thanks @AdityaHebballe  for working on this these search reliability improvements are really helpful. I noticed a couple of things while testing that might be worth considering.

Would it make sense to use one shared normalization function for cache keys and query comparisons?

```ts
const normalizeQuery = (query: string) =>
  query.normalize("NFKC").trim().toLowerCase().replace(/\s+/gu, " ");
```

This would keep casing and whitespace handling consistent across different languages and scripts. I reproduced an issue with `Batman` → `batman`, where the cached results disappeared.

Also, I may be missing something, but it looks like the TMDB unavailable warning is only rendered when partial results exist. If TMDB fails and no other provider returns results, wouldn’t the UI show “No matches” without the warning?

AdityaHebballe:
@aajsa 
Both reported issues are fixed.
      - Shared normalization now uses NFKC, trim, lowercase, and collapsed Unicode whitespace for cache keys and result-query comparison.
      - This fixes Batman → batman cached results disappearing.
      - TMDB unavailability is now detected for normal null TMDB responses, not only thrown errors.
      - The outage message renders independently of fallback results.
      - An empty TMDB outage no longer appears as misleading “No matches.”

aajsa (approved):

PR #953, opened 2026-07-19
fix: never leave layout:custom themes with zero navigation (#951)

### LINKED ISSUE #951: [Bug]: A community theme breaks Harbor, all the menus disappear, user settings is gone, I am only left with the home screen
### Before submitting

- [x] I searched existing issues and did not find a duplicate.

### Affected area

Interface, themes, or accessibility

### Harbor version

Latest (I can't check)

### Release channel

Beta

### Commit

_No response_

### Installation source

GitHub Releases

### System

Windows 11

### What happened?

Was looking at the new stuff through the latest beta, and I also looked to see if new themes are introduced, I selected one named dock or docker, I can't fully recall. Once I activated it, all the menus are gone, even user settings are gone, am left with only home page... now I cannot acccess any settings, I have tried pressing every key on the keyboard (no numpad), and in combinations with CTRL and Shift and both to see if I can open settings but no luck.

Image

### How to reproduce

1. Open Harbor
2. Enter Settings
3. Scroll until you see Themes
4. Browse community themes
5. Install the one called Dock(or docker), it's among the newer ones with 300 smth downloads
6. Activate it
7. Menus disappear, even the player settings

### Relevant logs

```shell

```

### Workaround

_No response_

### Screenshots, recordings, or supporting files

_No response_

AnayGarodia:
## Summary

Fixes #951 ("A community theme breaks Harbor, all the menus disappear, user settings is gone, I am only left with the home screen").

## Bug and root cause

`layout: "custom"` themes (the theme studio's "Custom" layout, and every community theme built the same way, e.g. the bundled Feishin/ElegantFin themes) render **100% of the app's navigation themselves** via injected HTML/CSS/JS (`src/components/custom-code-mount.tsx`). `App.tsx` intentionally renders none of its own nav chrome for that layout - only a back button and window controls (`src/App.tsx` around line 1022).

If the theme's script throws before it ever appends its nav element, or its markup simply never produces a working menu, `custom-code-mount.tsx` only does `console.warn(...)` on failure - invisible to a real user. The result: the user is left with **zero way to reach Settings or any other view**, exactly matching the reported symptom ("all the menus disappear, even user settings is gone").

I could not access the actual "Dock" community theme's source (it's hosted on `harbor.site`'s theme API, not in this repo, and the issue reporter wasn't certain of the exact name), so I could not do a byte-for-byte diff against it. Instead I reproduced the underlying failure class directly against `main`:

1. Opened the in-app Theme Studio, set layout to **Custom**.
2. In the Code tab, wrote a JS snippet that mirrors how a hand-rolled community "dock" theme typically builds its nav - `document.querySelector(...)` against a specific header/back-button selector, then a call on the result - simulating what happens when a theme's selector assumptions get invalidated by an app-side markup change.
3. Saved and activated the theme.

Console showed:
```
[harbor-theme-js] error: TypeError: Cannot read properties of null (reading 'getAttribute')
  at custom-code-mount.tsx:76:22
```
And the app was left with **no sidebar, no top bar, nothing** - just page content, unreachable Settings. Screenshot: the entire nav surface is gone, matching the issue's reported screenshot.

## Fix

Added `src/chrome/custom-layout-safety-net.tsx` (`CustomLayoutSafetyNet`), mounted in `App.tsx` alongside the existing custom-layout-only `FloatingBack`/`WindowControls`, gated the same way as every other layout's chrome (`!settingsTop && !playerActive && !pickerTop`).

It watches for any `[data-harbor-nav]` element - the attribute every built-in nav control (`Sidebar`, `TopDock`, ...) and every theme-authored nav control (the theme-studio chrome builder, the bundled Feishin/ElegantFin themes, and per the in-app theme cheat sheet, every community theme) is expected to use. If none exists a short moment after mount (and it keeps watching via `MutationObserver` for later changes), the theme's own navigation never showed up, so it renders a minimal built-in fallback menu covering the same destinations as `Sidebar` - reusing the shared `NAV_ITEMS` list and the exact same `useParental`/`ParentalPinModal` gating Sidebar already uses, so **the fallback cannot be used to bypass a parental PIN lock on Settings**.

Themes that render working navigation are completely unaffected - the check finds their nav and the component renders `null`.

The fallback's own stylesheet re-appends itself as the *last* `<style>` tag in `<head>` on every activation and forces `display`/`visibility`/`opacity`/`position`/`z-index`/`pointer-events` with `!important`, so it stays visible and clickable even against a theme that (deliberately or not) ships a broad `display: none !important` or `z-index` fight - the two failure patterns most likely to cause this class of bug.

## Verification

- `pnpm run typecheck` - clean.
- `pnpm run i18n:check` - clean (`en 4638, ar 3727, pt 4555 strings`); only added 2 new English keys, other locales don't need them per the script's own rules.
- `pnpm test` - 88/88 passing, including 4 new regression tests in `tests/custom-layout-safety-net.test.ts` (this repo has no DOM test harness - no jsdom/testing-library - so these follow the existing `tests/view-lifecycle.test.ts` precedent of pinning source-level invariants: the wiring in `App.tsx`, the `data-harbor-nav` detection, the `!important` defensive CSS, and the parental-PIN gating).
- `vp check` reports pre-existing formatting issues on this environment's toolchain across 819 files on a clean `main` checkout too (verified via `git stash`), unrelated to this change; none of the files this PR touches are in that list.
- Manually reproduced the bug and the fix live via a Vite dev server + headless Chromium: broken custom theme → no nav at all → fallback "Menu" button appears bottom-left → opens a full nav list → clicking Settings successfully navigates there. Then switched back to a healthy built-in theme and confirmed the fallback stays completely absent (no visual change for the 99% of themes that work correctly).
- No Rust changes, so `cargo check` doesn't apply here.

## Platform Impact

None - this is a plain React/TypeScript change with no native/platform-specific code paths.

## UI Changes

Only visible when a `layout: "custom"` theme fails to render its own navigation (previously: blank/broken app with no way to recover). Screenshots from manual verification:

**Before (broken theme, current `main`):** no nav surface at all.
**After (broken theme, this PR):** a small "Menu" button appears bottom-left; clicking it opens a full navigation list including Settings.
**Healthy theme (before and after):** identical, no fallback rendered.

## Checklist

- [x] I ran `vp check` for the changed files (pre-existing 819-file formatting baseline is unrelated, confirmed via `git stash` on clean `main`).
- [x] I ran `pnpm run typecheck` after TypeScript changes.
- [x] No Rust changes, so no Cargo checks apply.
- [x] Not platform-specific.
- [x] Preserved playback, navigation, and parental-PIN behavior (fallback nav reuses the same `useParental`/`ParentalPinModal` gate as `Sidebar`).
- [x] Added tests for the new behavior (`tests/custom-layout-safety-net.test.ts`).
- [x] No secrets/tokens/personal data in this PR.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


harborstremio:
hey sorry for the delays we will review this when we can also if you want to work on beta check out the beta-branch

harborstremio (approved):

PR #976, opened 2026-07-21
liquid glass in topbar, and fix poster-dock

O1Abdulrahman:
### Fix  poster-dock

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -2,9 +2,10 @@
 
 ## Task Completion Requirements
 
-- Run `vp check` for files changed by the task before considering it complete.
-- Run `vp run typecheck` after TypeScript changes.
+- Run `pnpm run check` (`vp check`) for files changed by the task before considering it complete.
+- Run `pnpm run typecheck` (`tsc -b --pretty false`) after TypeScript changes.
 - Run `cargo check --manifest-path src-tauri/Cargo.toml` after Rust changes.
+- Run `pnpm tauri:build:linux-system` (`tauri build --config src-tauri/tauri.linux-system.conf.json`) to build the full Linux binary.
 - Test platform-specific changes on the affected platform when possible.
 - Do not leave warnings or errors introduced by the task unresolved.
 - If a repository-wide check fails on untouched files, record the baseline and do not reformat or refactor unrelated files.
--- a/README.md
+++ b/README.md
@@ -9,8 +9,7 @@
 
 ### A Stremio Client Built for Adventure!
 
-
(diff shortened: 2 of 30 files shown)


AdityaHebballe:
image
Duplication of the title in the poster title. This is a regression.

- main: one throttled DOM transform write per affected card; the browser retargets the CSS transition.
- This PR: each pointer frame stops and creates a new Motion spring animation for every affected card.

I did change it back to a CSS transform based implementation. I have tuned it so it feels close to what you implemented while not feeling laggier on low-end systems as Motion spring animation can be expensive.

I have also made the speed of the poster dock animation customizable in the settings under Theme & Appearance.

Kept the liquid-glass theme direction, but cleaned up the implementation and poster dock behavior:
* Fixed the Liquid Glass compatibility/typecheck failures and removed the unused Motion dependency.
* Reworked poster dock to use CSS compositor transforms instead of creating/cancelling Motion springs on every pointer update. This keeps the same dock scale/lift/spread while reducing hover overhead.
* Limited reset work to posters that were actually dock-active, and animate the poster visual rather than its whole card to avoid title artifacts on hover.
* Added a persisted Poster Dock animation-speed setting (250–1500ms, default/reset: 760ms), with safe loading/clamping.
* Added a full-width live six-poster preview in Theme settings so users can tune the speed before using it in real rows.

Frontend checks, typecheck, and the Liquid Glass test pass.

AdityaHebballe (approved):

O1Abdulrahman:
> image Duplication of the title in the poster title. This is a regression.
> * main: one throttled DOM transform write per affected card; the browser retargets the CSS transition.
> * This PR: each pointer frame stops and creates a new Motion spring animation for every affected card.
> 
> I did change it back to a CSS transform based implementation. I have tuned it so it feels close to what you implemented while not feeling laggier on low-end systems as Motion spring animation can be expensive.
> 
> I have also made the speed of the poster dock animation customizable in the settings under Theme & Appearance.
> 
> Kept the liquid-glass theme direction, but cleaned up the implementation and poster dock behavior:
> 
> * Fixed the Liquid Glass compatibility/typecheck failures and removed the unused Motion dependency.
> * Reworked poster dock to use CSS compositor transforms instead of creating/cancelling Motion springs on every pointer update. This keeps the same dock scale/lift/spread while reducing hover overhead.
> * Limited reset work to posters that were actually dock-active, and animate the poster visual rather than its whole card to avoid title artifacts on hover.
> * Added a persisted Poster Dock animation-speed setting (250–1500ms, default/reset: 760ms), with safe loading/clamping.
> * Added a full-width live six-poster preview in Theme settings so users can tune the speed before using it in real rows.
> 
> Frontend checks, typecheck, and the Liquid Glass test pass.

thanks 🙏

PR #989, opened 2026-07-22
Fix NVIDIA WebKit rendering on Wayland

encomjp:
## What changed

I was running into three different rendering problems with the existing NVIDIA workaround on Wayland:

- forcing X11 made the embedded mpv video blank
- using native Wayland with WebKitGTK's DMA-BUF renderer disabled left stale transparent UI pixels over the video
- enabling DMA-BUF without an NVIDIA workaround crashed with a Wayland protocol error 71

This keeps the existing DMA-BUF fallback for NVIDIA on X11, but changes the Wayland path to stay native and set `__NV_DISABLE_EXPLICIT_SYNC=1`. That allows WebKitGTK's DMA-BUF renderer to remain enabled, so transparent overlays composite and clear correctly, while avoiding the protocol error.

Environment variables explicitly provided by the user are still respected.

## Test setup

- CachyOS rolling
- KDE Plasma / KWin 6.7.3 on Wayland
- Linux 7.1.4-1-cachyos-bore
- NVIDIA GeForce RTX 5080 Laptop GPU
- NVIDIA driver 610.43.03
- WebKitGTK 2.52.5
- GTK 3.24.52

## Testing

- Built the Linux system package successfully
- Launched Harbor through NVIDIA PRIME render offload
- Confirmed embedded mpv video renders instead of staying blank
- Confirmed subtitles, timeline, tooltips, and menus clear without stale pixels
- Confirmed the Wayland protocol error no longer occurs

PR #990, opened 2026-07-22
Add native Linux SVP support

encomjp:
## What changed

The Anime tweaks panel was already available on Linux, but its SVP backend only searched Windows install paths and required `SVPManager.exe`, `VSScript.dll`, and SVPFlow DLLs. This adds a native Linux path while preserving the existing Windows behavior.

- Detects SVP from environment overrides, PATH, user installs, and `/opt/svp*`
- Finds matching native `libsvpflow1.so` and `libsvpflow2.so` plugins
- Detects modern, legacy, versioned, and multiarch VSScript library layouts
- Configures modern VapourSynth installations at startup with a bounded timeout
- Verifies the libmpv linked into Harbor accepts the VapourSynth filter
- Generates the script with exact native plugin paths and avoids reloading auto-loaded namespaces
- Starts native SVP Manager when available
- Uses a labeled mpv filter so failure cleanup does not remove unrelated video filters
- Disables SVP cleanly when the Flatpak sandbox or linked libmpv cannot support it
- Enables built-in smooth motion for embedded Linux playback instead of skipping it
- Makes Anime4K require a complete shader download and keeps the global off toggle authoritative

## Test setup

- CachyOS rolling
- KDE Plasma / KWin 6.7.3 on Wayland
- Linux 7.1.4-1-cachyos-bore
- NVIDIA GeForce RTX 5080 Laptop GPU, driver 610.43.03
- SVP 4.7.305
- VapourSynth R77 / API R4.2
- mpv 0.41.0-878-g94335ab87 with VapourSynth enabled

The current stable mpv 0.41 package predates mpv's April 2026 VSScript R74 loader update, so I used current mpv git for the runtime test. Harbor now reports the linked library as unsupported instead of enabling SVP when its VapourSynth filter is unavailable.

## Testing

- Production frontend build
- 85 frontend tests
- 59 Rust tests
- Native SVPFlow smoke test converting 24 fps to 60 fps through both Linux plugins
- Embedded Harbor playback with the labeled VapourSynth filter initialized successfully
- Interactive playback checks for SVP, Anime4K, and built-in smooth motion
- NVIDIA Wayland playback remains working after rebasing on #989


AdityaHebballe:
Thank you for bringing more features into linux! I will merge this.

AdityaHebballe (approved):

PR #1001, opened 2026-07-23
added Liquid Glass to the volume HUD and the window controls, fixed t…

O1Abdulrahman:
### I added Liquid Glass to the volume HUD and the window controls, including the minimize, maximize, restore, and close buttons.
Liquid Glass will only be rendered on the HTML player. It will not be shown when the MPV player is active, as the other approaches I tried were not practical.

### I also fixed the Top Bar settings behavior. Previously, disabling Liquid Glass did not properly remove the effect.

### DIFF
--- a/src/chrome/royal-topbar.tsx
+++ b/src/chrome/royal-topbar.tsx
@@ -1,4 +1,11 @@
-import { useEffect, useLayoutEffect, useRef, useState, type ReactNode } from "react";
+import {
+  useEffect,
+  useLayoutEffect,
+  useRef,
+  useState,
+  type ComponentProps,
+  type ReactNode,
+} from "react";
 import { LogIn, LogOut, Pencil, Search, Settings as SettingsLucide, Users } from "lucide-react";
 import { createPortal } from "react-dom";
 import { HarborMark } from "@/components/icons/harbor-mark";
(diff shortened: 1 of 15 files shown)


AdityaHebballe:
> It will not be shown when the MPV player is active, as the other approaches I tried were not practical.

But it does though. Is this intended?
StageOverlays renders VolumeIndicator for both engines. It simply passes allowBoost={engine === "mpv"}. The updated VolumeIndicator always renders ThreeLiquidGlassSurface, so mpv gets the new Glass HUD too.

It does look good but there might be a few bugs that need fixing. Let's have it on both mpv and HTML5

O1Abdulrahman:
> > It will not be shown when the MPV player is active, as the other approaches I tried were not practical.
> 
> But it does though. Is this intended? StageOverlays renders VolumeIndicator for both engines. It simply passes allowBoost={engine === "mpv"}. The updated VolumeIndicator always renders ThreeLiquidGlassSurface, so mpv gets the new Glass HUD too.
> 
> It does look good but there might be a few bugs that need fixing. Let's have it on both mpv and HTML5

the 

> But it does though. Is this intended? StageOverlays renders VolumeIndicator for both engines. It simply passes allowBoost={engine === "mpv"}. The updated VolumeIndicator always renders ThreeLiquidGlassSurface, so mpv gets the new Glass HUD too.
> 
> It does look good but there might be a few bugs that need fixing. Let's have it on both mpv and HTML5

the backdropFilter does not actually interact with or blur the MPV video surface. What you’re seeing is mainly the transparent tint. It looks good but not the same as HTML. I’ll try a few other approaches to see if I can make it work more efficiently with MPV.

AdityaHebballe:
@OakArbum I've fixed the bugs for liquid glass on MPV. Can you check? If it's fine I'll merge this. I did not test using HTML5 as the stream does not play for me.

AdityaHebballe (dismissed):

AdityaHebballe (dismissed):

AdityaHebballe (dismissed):

AdityaHebballe:
Fixed a lot of other bugs, like volume card not hiding when player controls are active. Delayed blur effects for default and enhanced renderer on play/pause, back buttons and volume card.

AdityaHebballe (approved):

O1Abdulrahman:
> Fixed a lot of other bugs, like volume card not hiding when player controls are active. Delayed blur effects for default and enhanced renderer on play/pause, back buttons and volume card.

Thank you for your work on this. However, this is how it looks in HTML, and it used to look like this before.

The blur/refraction effect we created with `backdrop-filter` cannot detect or sample the content behind the MPV player. This is because MPV is rendered through a separate native video surface outside the WebView’s normal HTML/CSS rendering layer. As a result, `backdrop-filter` can only process elements rendered inside the WebView, while the MPV video appears as an external layer that CSS cannot access. This is why the effect works correctly with the HTML5 player but not with MPV.

Screenshot 2026-07-23 154556

Screenshot 2026-07-21 135135

Screenshot 2026-07-21 122237

For example, this image was captured from the MPV player. As you can see, the Liquid Glass effect reacts correctly to the Seek Preview Thumbnail, but it does not react to the live MPV video behind the Skip Intro button.

Screenshot 2026-07-23 155721

It seems there is no straightforward solution for this, except through relatively complex workarounds. Thank you for all the effort you’ve put into this. I’m truly grateful and really appreciate your work.

PR #1002, opened 2026-07-23
fix(detail): move season picker beside episodes

AdityaHebballe:
## Summary

Moves the season, arc, franchise, and TVDB order selectors beside the Episodes heading instead of placing them at the far right of the episode header.

This applies to the TMDB series, anime, and Cinemeta fallback episode views.

Before:
Screenshot 2026-07-23 at 12 49 06 PM
After:
Screenshot 2026-07-23 at 12 48 43 PM

This gives a cleaner look and makes the episode bar feel less cluttered 
## Verification

- `pnpm run check`
- `pnpm run typecheck`


JJDizz1L (approved):
Looks good.

PR #1003, opened 2026-07-23
fix: keep first CW item visible after playback reorder

Thunderhawkk:
## Summary

Fix Continue Watching item "disappearing" from home screen after playback by checking first-child visibility on every Row layout effect.

## Why

When a CW item is played, saveLocalCw updates and reorders the items — the most recently played moves to position 1. The Row component restores the saved scroll position, which hasn't accounted for the reorder, pushing the now-first item out of view to the left. Previously the check only ran on the initial mount; it now runs on every layout effect, resetting scroll to 0 when the first child is scrolled out of view.

## Verification

pnpm run check  
pnpm run typecheck  
Manual: play a CW item from position 2, exit, confirm item now visible at position 1

## Platform Impact

None — the Row component is shared and scroll behavior is consistent across all platforms.

## UI Changes

N/A

## Checklist

### DIFF
--- a/src/components/row.tsx
+++ b/src/components/row.tsx
@@ -210,7 +210,17 @@ export function Row({
       const target = n != null && n > 0 && max > 0 ? Math.min(n, max) : 0;
       if (readPos(trackEl) !== target) writePos(trackEl, target);
       restoredRef.current = true;
-      return;
+    }
+    if (childCount > 0) {
+      const first = trackEl.firstElementChild as HTMLElement | null;
+      if (first) {
+        const tr = trackEl.getBoundingClientRect();
+        const fr = first.getBoundingClientRect();
+        if (fr.right <= tr.left + 4) {
+          writePos(trackEl, 0);
+          return;
+        }
+      }
     }
     if (!userInteractedRef.current && readPos(trackEl) !== 0) {
       writePos(trackEl, 0);
--- a/src/components/row.tsx
+++ b/src/components/row.tsx
@@ -210,17 +210,7 @@ export function Row({
       const target = n != null && n > 0 && max > 0 ? Math.min(n, max) : 0;
       if (readPos(trackEl) !== target) writePos(trackEl, target);
       restoredRef.current = true;
-    }
-    if (childCount > 0) {
-      const first = trackEl.firstElementChild as HTMLElement | null;
-      if (first) {
-        const tr = trackEl.getBoundingClientRect();
-        const fr = first.getBoundingClientRect();
-        if (fr.right <= tr.left + 4) {
-          writePos(trackEl, 0);
-          return;
-        }
-      }
+      return;
     }
     if (!userInteractedRef.current && readPos(trackEl) !== 0) {
       writePos(trackEl, 0);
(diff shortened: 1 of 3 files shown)


AdityaHebballe:
- Saved row positions are effectively broken. Removes the return after restoring
    scrollKey position. On initial restore, if the first card is fully offscreen it immediately
    resets to 0; if it is only partly offscreen, the existing fallback below also resets it to
    0 because userInteractedRef is still false. This affects every remembered row, not only
    Continue Watching.

  - It can snap a user back to the start during live updates. The new geometry check runs
    before the userInteractedRef guard, so any row whose first item is offscreen can jump back
    to zero when its children update—even if the user intentionally scrolled there.

  - RTL is not handled. fr.right <= tr.left + 4 only detects a card disappearing to the left.
    The beta Row explicitly supports RTL scrolling; in RTL the first item can disappear beyond
    the right edge instead.

  - Wrong scope / unnecessary layout work. Row is used by home, catalog, discover, anime,
    collections, etc. Calling getBoundingClientRect() on every layout effect across all rails
    is not a Continue Watching-specific solution.
    
 Did you observe this only for continue watching? if so we need to keep this more tightly scoped to continue watching row

Thunderhawkk:
> * Saved row positions are effectively broken. Removes the return after restoring
>   scrollKey position. On initial restore, if the first card is fully offscreen it immediately
>   resets to 0; if it is only partly offscreen, the existing fallback below also resets it to
>   0 because userInteractedRef is still false. This affects every remembered row, not only
>   Continue Watching.
> * It can snap a user back to the start during live updates. The new geometry check runs
>   before the userInteractedRef guard, so any row whose first item is offscreen can jump back
>   to zero when its children update—even if the user intentionally scrolled there.
> * RTL is not handled. fr.right <= tr.left + 4 only detects a card disappearing to the left.
>   The beta Row explicitly supports RTL scrolling; in RTL the first item can disappear beyond
>   the right edge instead.
> * Wrong scope / unnecessary layout work. Row is used by home, catalog, discover, anime,
>   collections, etc. Calling getBoundingClientRect() on every layout effect across all rails
>   is not a Continue Watching-specific solution.
> 
> Did you observe this only for continue watching? if so we need to keep this more tightly scoped to continue watching row

yeah, I'll do that, I just tested it on CW since I was working on it

Thunderhawkk:
hey @AdityaHebballe can you check with the new commit? I reverted the change from the 1st commit and limited the scope to CW Section only, I tested it on my side for other rows as well and it seems to be working fine, let me know if you find any problem

AdityaHebballe:
- cwScrollKey includes first.state?.lastWatched. That value changes during playback autosaves, so it can remount the Continue Watching row repeatedly and creates a new in-memory row-scroll key each time. The key should only reflect the reorder condition (e.g. items[0]?._id) while keeping scrollKey="home:cw" stable.
- The PR also changes use-resume-autosave.ts to save local Continue Watching for all streams. That is unrelated to the stated scroll fix and substantially changes playback behavior.

Can you please check this?

Thunderhawkk:
ok I will look into it, this CW stuff is so tricky

AdityaHebballe:
Yeah lol, no worries take your time

Thunderhawkk:
Made some changes please check them out @AdityaHebballe ran a few tests all passed

AdityaHebballe:
Looks good! Thank you for your contribution!

Issue #1007, opened 2026-07-23
Proposal: official Linux package maintenance


AdityaHebballe:
I would like to maintain Harbor’s Linux packages officially: AUR, Debian/Ubuntu, Fedora/openSUSE, Flatpak, and AppImage(after a few issues with it are ironed out).

The goal is simple: Linux users install Harbor once through their normal package manager and receive updates normally afterward.

## What I need from Harbor

1. Confirm that @harborstremio  wants these to be official Linux distribution channels.
2. Create a dedicated repository, ideally `harborstremio/linux-packages`.
3. Grant `@AdityaHebballe` Admin access to that repository.
4. Approve an official Harbor Flatpak submission to Flathub

I do **not** need Harbor organization Owner access or a separate Harbor GitHub account.

## What I will handle

- package builds and release automation;
- package repositories and updates;
- AUR maintenance;
- Flatpak packaging and maintenance;
- install documentation and release checks.

I will keep stable and beta packages clearly separated and make sure package publishing is tied to exact Harbor releases or commits.

We can discuss further in Discord,etc or continue the conversation here as well. 

Thank you!

harborstremio:
> I would like to maintain Harbor’s Linux packages officially: AUR, Debian/Ubuntu, Fedora/openSUSE, Flatpak, and AppImage(after a few issues with it are ironed out).
> 
> The goal is simple: Linux users install Harbor once through their normal package manager and receive updates normally afterward.
> 
> ## What I need from Harbor
> 
> 1. Confirm that @harborstremio  wants these to be official Linux distribution channels.
> 2. Create a dedicated repository, ideally `harborstremio/linux-packages`.
> 3. Grant `@AdityaHebballe` Admin access to that repository.
> 4. Approve an official Harbor Flatpak submission to Flathub
> 
> I do **not** need Harbor organization Owner access or a separate Harbor GitHub account.
> 
> ## What I will handle
> 
> - package builds and release automation;
> - package repositories and updates;
> - AUR maintenance;
> - Flatpak packaging and maintenance;
> - install documentation and release checks.
> 
> I will keep stable and beta packages clearly separated and make sure package publishing is tied to exact Harbor releases or commits.
> 
> We can discuss further in Discord,etc or continue the conversation here as well. 
> 
> Thank you!
> 

[...]

PR #1010, opened 2026-07-23
fix: defer Linux updates to package managers

AdityaHebballe:
## Summary
- hide the in-app Updates section on Linux
- disable the global update watcher and update prompt on Linux

Linux packages are updated by their package manager rather than Harbor's self-updater.

## Verification

- pnpm run check 
- pnpm run typecheck
- pnpm tauri:build:linux-system

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,7 +18,8 @@ Offering things like a native player, a stream ranking engine, Wikidata, watch p
 [![Tauri][badge-tauri]][tauri] &nbsp;
 [![React][badge-react]][react] &nbsp;
 [![Rust core][badge-rust]][rust] &nbsp;
-[![Platforms][badge-platforms]][releases]
+[![Platforms][badge-platforms]][releases] &nbsp;
+[![Linux packages][badge-linux-packages]][linux-packages]
 
 <br/>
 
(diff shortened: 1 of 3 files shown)


JJDizz1L (dismissed):
Makes sense as Linux usually updates through repositories.

JJDizz1L (approved):

AdityaHebballe:
@harborstremio Added reference to Linux Unofficial Builds repo in README. If there are any issues related to Linux packaging I will handle it.

AUR, RPM, DEB, and Flatpak are ready now. There are few issues with Appimage and Tauri, so we need to wait for Tauri to fix it or work around it ourselves. 

These workflows can be shifted to `harborstremio/linux-packages` as discussed in #1007, then the package can be updated from package manager for RPM, DEB and Flatpak.

I will submit the Flatpak to Flathub after the repo is created.

JJDizz1L:
@AdityaHebballe Submit Harbor to this as well, please.
Website: https://flatpark.org/
GitHub Project: https://github.com/flatpark/flatpark

AdityaHebballe:
Sure @JJDizz1L, I will do that

PR #1013, opened 2026-07-23
feat: timer mode for anime CW end behavior + fix rewatch advance with AniList/SIMKL completed status

Thunderhawkk:
## Summary

Add animeCwEnd: "timer" mode for Continue Watching — when an anime episode ends and the next hasn't aired yet, keep the item with a live countdown to the air date instead of removing it. Merge and integrate with beta-branch's cwHideCaughtUp feature.

## Why

Timer mode lets users see exactly when the next episode drops without leaving the CW row, while hide mode (default, now covered by cwHideCaughtUp) removes finished items cleanly.

The advance predicate fixes prevent external services (AniList, SIMKL) marking a series as "completed" from blocking rewatch advance — simklCompleted is scoped to episodes before the current one, and nextUnwatchedAfter uses a local-only predicate so every episode after the current resume point is treated as unwatchable regardless of external state.

## Verification

- pnpm tsc --noEmit passes
- node --test — 67 pass, 2 pre-existing failures
- Timer mode: finish latest aired episode → item stays with countdown → refreshes every 60s
- Hide/cwHideCaughtUp: finish latest aired episode → item removed from CW
- Rewatch with AniList "completed": item advances to next episode instead of being dropped

## Platform Impact

None — frontend-only.

## UI Changes

"Timer" option in the "When the latest episode ends" Segmented control in library settings.

## Checklist

- [ x I removed secrets, tokens, private URLs, and personal data from logs and screenshots.


Thunderhawkk:
@AdityaHebballe can you check this out if you get the time and let me know if it conflicts with something else or if there are problems, only the latest commit really matters since I was working on a older version before the merge, after merge it was already present, made the few changes that actually mattered in the newest commit

AdityaHebballe:
- Timer cards never advance when the episode airs. The hook creates a waitingForAir
    card, but nothing reruns useCwAdvance at nextAirDate. The card’s own interval only changes
    the label to “Airing now,” so it can remain stuck there indefinitely until an unrelated
    state update occurs. Add a scheduled refresh/re-evaluation at the air time.

    https://github.com/harborstremio/harbor/blob/b86cacc60b357965e2c097b4dde9eab6d15922c6/src/views/home/hooks/use-cw-advance.ts#L239

  - Changing Hide ↔ Timer does not apply immediately. animeCwEnd is used inside the
    effect but omitted from its dependency list. Toggling the setting leaves the previously
    calculated CW state in place until some other dependency changes. Add animeCwEnd to the
    dependencies. 
    https://github.com/harborstremio/harbor/blob/b86cacc60b357965e2c097b4dde9eab6d15922c6/src/views/home/hooks/use-cw-advance.ts#L284

Thunderhawkk:
made those changes, please review it, thank you @AdityaHebballe

PR #1029, opened 2026-07-25
fix(macos): wire up embedded HDR so the EDR path actually activates

### LINKED ISSUE #361: HDR USERS
Please use this thread to report back your feedbacks (remind us of your resolution and monitor (is it external etc) ) 

We closed the others to condense it into one

d64s54csd9-source:
Fixes #361

## Problem

Embedded playback on macOS never produced true HDR. The EDR implementation in
`mpv_render_mac.rs` and `apply_mac_edr()` in `mpv.rs` were fully written but
unreachable: `use-player-bridge.ts` hardcoded `macEdr: false`, so
`apply_mac_edr()` was never called.

`settings.playerMacEdr` exists in the settings types, but nothing on macOS ever
writes it — `HdrModePicker`, the only place that sets the HDR mode flags, is
rendered behind `isWindowsDesktop()` and never appears on macOS.

## Fix

Derive the flag from the two toggles the macOS UI actually exposes ("Embed mpv
inside Harbor window" and "HDR-to-SDR tonemapping") instead of a stored value
no macOS UI writes:

```ts
macEdr: isMacDesktop() && embedActive && !settings.playerHdrToSdr,
```

## Verification

M4 MacBook Air, macOS 26.0, mpv 0.41.0, `vo=libmpv`, 4K DV/HDR10 source
(`dolbyvision/bt.2020/pq`).

Before: `target-trc` / `target-prim` never appear in the mpv log; the render
pass ends in `// HDR tone mapping`, `pow(color.rgb, vec3(1.0/2.2))`,
`Dither to 8`.

After:

```
Set property: icc-profile-auto="no" -> 1
Set property: target-prim="bt.2020" -> 1
Set property: target-trc="pq"       -> 1
Set property: target-peak="auto"    -> 1
```

## Follow-ups this exposes (not fixed here)

**1. Window colorspace vs. mpv output mismatch.** With EDR active the video is
visibly crushed and desaturated. mpv emits PQ / BT.2020 while
`mpv_render_mac.rs` tags the window `kCGColorSpaceExtendedLinearDisplayP3` — a
*linear* P3 space. PQ code values read as linear light explains the result.
`set_hdr_active(active, _bt2020)` also ignores its `bt2020` argument, so
primaries are never matched.

I tested tagging the window `kCGColorSpaceITUR_2100_PQ` instead: **the video
then renders correctly in true HDR**. However `NSWindow.setColorSpace` applies
to the whole window, so the sRGB WebView UI on top is re-interpreted and turns
dark and colour-shifted. Not submitted for that reason.

This looks like the same problem the Windows path already solves with
`hdr_overlay.rs` + `use-hdr-stage.ts` (video plane in its own colorspace,
controls in a separate overlay window). Extending that HDR-stage architecture
to macOS — or setting the colorspace per-layer rather than per-window — seems
to be the missing piece.

**2. Output stays 8-bit.** `Dither to 8` even with EDR active.

**3. `hdr-compute-peak` unavailable.** macOS caps OpenGL at 4.1, so
`compute shaders=0, SSBO=0` and mpv disables peak detection on this path.

**4. Periodic full-resolution software screenshots.** `mpv_screenshot_data_url`
sets `screenshot-sw=yes`, and the continue-watching hook fires every 12s. On
the embedded path each capture converts a 3840x1920 p010 frame to rgb24 with
zimg across 9 CPU threads — a visible hitch during 4K HDR playback on fanless
hardware. Decoding also runs `videotoolbox-copy` rather than direct interop.

**5. Unrelated packaging note.** `tauri-build.yml` produces a macOS app that
cannot launch: `fetch-libmpv.mjs` skips macOS ("mac and linux link system
libmpv"), and nothing rewrites the absolute Homebrew path baked into the binary
or bundles `Contents/Frameworks`, unlike the official DMGs.


ricardosousa191:
Hello, i've benn testing this and the hdr is not working properly 
Captura de ecrã 2026-07-25, às 11 21 12

When playing a sdr after watching a hdr media it starts blinking until going to window and fullscreen again

https://github.com/user-attachments/assets/be6d2e1f-2f38-4ec4-81de-db633e3ea254

Using pq doesn't calibrate the hdr correctly, edr is linear

d64s54csd9-source:
@ricardosousa191 
I know that I mentioned these problems in the "Follow-ups (this exposes, not fixed here)." This issue is not fixed yet. Sometimes it works properly, and sometimes it doesn't.

ricardosousa191:
@d64s54csd9-source I understand the follow-up, but you can't call this a fix in the title (fixes 361 too) and opening a pull request like this to something not working. Most people using this app doesn't have mac so many would think it would fix most of it even with the follow-up.
Edit: Turning on the edr in the code doesn't fix the problem, it introduces more problems because you need to calibrate everything and do all the edge cases.

ricardosousa191:
The new commit made a lot of regressions
1. Now there is a line in the video player
2. The player doesnt show the media controls and subtitles when in fullscreen
3. Image flickers until fullscreen is applied

Flickering image warning 
https://github.com/user-attachments/assets/08e1c60c-6b2d-4be8-b7ba-ee3dc833ecb1

You are leaving the scope and introducing new issues. Please when doing a commit use conventional commits.

JJDizz1L (approved):
HDR for Mac =]

ricardosousa191:
@JJDizz1L How was this approved lol, its introducing so much regressions to mac player. 
Do you guys do testing before accepting anything? In my videos the player is flicking so fast in hdr that would give some epiletic guy a damage.
Flickering image warning


https://github.com/user-attachments/assets/e07e16c1-980e-47cb-885f-f0fe46643b0d




@AdityaHebballe Could you take a look at this? Thanks

JJDizz1L:
@ricardosousa191 Relax. It's not merged. And it's being tested.

ricardosousa191:
@JJDizz1L The point is not the merge, it is being approved, you cant approve something just because of the title and being worked on. You only approve when it meets the standards. And who is doing the testing because you need to document that in the pr.

JJDizz1L:
@ricardosousa191 Send a PR of your own improving your conflict of interests. This back and forth isn't helpful.

ricardosousa191:
@JJDizz1L Why would i open a pr? I have came to the conclusion that mac edr is not a simple thing and did a lot of work overnight. I wont open a pr to something i dont feel its good enought and would create more problems.

Im just here testing the pr and saying whats wrong.
If there is a problem with the pr i will say it either way, im here for that.

I just want to point out that having a concern about a regression that could potentially harm people with epilepsy isn't a conflict of interest.

AdityaHebballe:
@ricardosousa191 @JJDizz1L It's ok guys there's no emergency to merge this PR. Let's thoroughly test this and make changes.

I will take a look to tomorrow. Let's make sure we merge some quality code.

ricardosousa191:
@AdityaHebballe Thanks!

AdityaHebballe:
@ricardosousa191 I can confirm that this approach is extremely buggy, I got multiple line artifacts on the video but could not replicate the flickering. Nonetheless we can't merge this yet.

I think we need to revert the child-window commit. It is the source of the seams, flicker, overlay regressions.

We might need to hold off on this feature until we find a way that's more stable.

ricardosousa191:
@AdityaHebballe Hey thanks, i did some coding about this feature in my own build and can confirm that edr/hdr libmpv is really buggy and hard to implement on mac.

I changed a lot less code and calibrated everything and being honest it works but has 3 main issues in my build
1. When you watch a hdr video, you need to put the media in fullscreen first for it to work in fullscreen and window.
2. When switching from hdr to sdr it causes a huge oversaturation problem, where i tried to change the process of when to load the config and how much ms for it, no result tho
3. Somehow during my coding i introduced a bug that i cant fix, the video starts to have lines in the borders (only hdr media)

 I think the main problem for this feature in macos is the usage of opengl in libmpv and not metal, edr suits better with the metal layer made by apple than opengl.

For this to work perfectly we needed to change the layer for mac to metal and that gives too much of headache.

PR #1031, opened 2026-07-25
feat(streams): persist active stream filter

aajsa:
## Included
- Persist the selected saved stream filter.
- Apply it to manual and automatic picks with safe fallback when nothing matches.
- Bypass the preference for Together host matching and preserve previous-source lookup.
- Add settings controls and active/fallback feedback.

## Verification
- focused stream-filter tests — 3/3
- full TypeScript tests — 88/88
- `pnpm run typecheck`
- full Tauri production build
- GitHub Frontend and Native checks passed


AdityaHebballe:
Does the active stream filter intentionally allow remembered and season-locked streams to override it during autoplay/resume?

PR #1045, opened 2026-07-26
feat: port background download manager from beta

aajsa:
## Included
- Port the beta background auto-download manager for series and anime.
- Resolve metadata through the saved addon origin, Anime Kitsu, Cinemeta, then compatible installed metadata addons.
- Add season and episode download actions with stream resolution.
- Expand Downloads with active torrents, pause/resume controls, save location, and streaming state.
- Preserve torrent lifecycle settings and expose native list/pause/resume commands.

## Reference
- Ported from `beta-branch` while keeping the change isolated for parity review.

## Verification
- `pnpm run typecheck`
- `pnpm test` — 92/92
- `pnpm run i18n:check`
- changed-file check passed with no warnings or errors
- GitHub Frontend and Native checks passed


harborstremio:
Hey @aajsa do you want to port these to main manually or do you think its better to just ensure beta has all current PRs/commits and then merge over the main with the current beta like we used to? Think that would be more efficient in the long run just keep beta up to date when its stable > gets moved to main rinse and repeat, If you still want to port it over though that is totally okay but eventually we will just do a blanked merge over cause beta has a ton more than main, we would maybe do a legacy harbor branch which is this non profile/social /sync version + some beta features we port over. let me know your thoughts

I think we are gonna do the following. Main stays the legacy and then we do stable branch/ beta branch etc releases independently so your work is preserved

AdityaHebballe:
@aajsa this looks done can we merge?

AdityaHebballe (approved):

aajsa:
@harborstremio
That makes sense. After working on the manga port and comparing both branches more closely, I realized that manually porting every beta feature into the current `main` probably isn’t the most efficient long-term approach.

`beta-branch` already contains much more of the complete Harbor experience, while the current `main` has a stronger technical foundation in some areas, including Vite Plus, TanStack Router, Query and Virtual, newer dependencies, better testing and i18n validation, and some important security and platform fixes.

So I agree with keeping the current `main` as the legacy version and maintaining stable and beta independently for now. We can focus on keeping `beta-branch` updated with the important fixes and PRs from `main`, then gradually bring the stronger architecture and tooling into the beta codebase later.

The manga PR is already complete and contains useful fixes and improvements, so you can still merge it into the legacy `main`. For future work, though, I think focusing on `beta-branch` will be the more efficient direction.

PR #1064, opened 2026-07-27
fix(linux): bundle AppImage media framework

danielcadev:
## Summary

Bundle Tauri's Linux media framework in AppImages and install the GStreamer plugin sets used by the Linux packaging workflow.

## Why

The AppImage did not include the GStreamer runtime plugins WebKitGTK needs for media playback. System packages can use distribution dependencies, but the AppImage needs these plugins bundled to provide the same playback and audio support.

## Verification

- Ran `vp fmt --check` on the changed YAML and JSON files.
- Ran `vp check --no-fmt` with 0 errors.
- Ran the TypeScript and Cargo checks successfully.
- Built the Linux release binary, DEB, and RPM.
- Built and extracted a Fedora 44 validation AppImage.
- Confirmed the AppImage contains 241 GStreamer plugins, including `libgstapp.so`, `libgstautodetect.so`, and `gst-plugin-scanner`.
- Manually verified video playback, TrueHD audio, navigation, and seek-preview behavior on Fedora KDE Wayland.
- Confirmed JSON and workflow YAML parsing and `git diff --check`.

The local Fedora AppImage validation used `NO_STRIP=1` because the `linuxdeploy` binary bundled by the local Tauri tooling cannot strip Fedora's newer `.relr.dyn` binaries. This was only a local packaging workaround and is not part of the patch.

## Platform Impact

Linux AppImage only. DEB and RPM dependency behavior remains unchanged. Windows and macOS are unaffected.

## UI Changes

None.

## Checklist

- [x] I ran `vp run typecheck` after TypeScript changes. No TypeScript files changed; the equivalent TypeScript check passed.
- [x] I ran the relevant Cargo checks after Rust changes. No Rust files changed; `cargo check --offline --locked` passed.
- [x] I tested affected platforms when the change is platform-specific.
- [x] I used the closest applicable regression verification for this packaging-only change: artifact extraction plus runtime playback testing.

### DIFF
--- a/.github/workflows/tauri-build.yml
+++ b/.github/workflows/tauri-build.yml
@@ -52,6 +52,10 @@ jobs:
             libasound2-dev \
             libmpv-dev \
             libgtk-3-dev \
+            gstreamer1.0-plugins-base \
+            gstreamer1.0-plugins-good \
+            gstreamer1.0-plugins-bad \
+            gstreamer1.0-libav \
             patchelf
 
       - name: Install Mac deps
--- a/src-tauri/tauri.conf.json
+++ b/src-tauri/tauri.conf.json
@@ -65,6 +65,9 @@
       "icons/icon.ico"
     ],
     "linux": {
+      "appimage": {
+        "bundleMediaFramework": true
+      },
       "deb": {
         "depends": [
           "libmpv2 | libmpv1",


AdityaHebballe:
Appimage was intentionally left unfinished as it appears to be buggy on several distros. I will test appimage on multiple distros and add anything necessary to this PR and merge it.

Thank you for your contribution!

danielcadev:
Small follow-up on the openSUSE test:
I built the merged commit on Ubuntu 22.04, matching the release workflow. The resulting AppImage contains 239 GStreamer plugins, including playback, PulseAudio, appsrc/appsink, and automatic audio output support.
I also tested Harbor on openSUSE Tumbleweed 20260724 with those media plugins available, and video and audio playback worked.
I could not complete a clean end-to-end test of that exact Ubuntu-built AppImage inside the container because WebKitGTK failed during EGL initialization before displaying the UI. This appears specific to the container graphics environment, but I do not want to claim a complete openSUSE AppImage pass.
The packaging change looks correct based on the artifact contents and playback test, while the exact cross-distribution AppImage validation remains pending.

AdityaHebballe:
I ran into the same EGL issue and initially held off distributing the AppImage because of it, although I think it's still best to distribute it labelled as experimental and collect feedback.

What ended up working for me was building the AppImage first, then removing the bundled GL/EGL/GBM/Wayland graphics libraries, along with the bundled TLS/HTTP libraries, before repacking it.

That way, WebKitGTK, Mesa, and the media transport use the host distribution's compatible libraries, while Harbor, libmpv, FFmpeg, the codecs, and the GStreamer plugins remain bundled inside the AppImage.

From what I could tell, this seems to be a fairly common AppImage/linuxdeploy interoperability issue rather than something specific to Harbor. I tested the rebuilt AppImage locally, and the UI loads correctly, with both video and audio playback working.

I'm on Arch Linux and am looking for alternatives so that I don't have to do this, but I've had no luck so far.

I'm currently trying sharun:
https://github.com/harborstremio-linux/harbor-linux-builds/blob/main/.github/workflows/test-portable-appimage.yml

However it keeps freezing and it's still a draft PR, so I can't consider it for a production release.

You're welcome to have a look at my packaging workflow as well:
https://github.com/harborstremio-linux/harbor-linux-builds/blob/main/.github/workflows/build-appimage.yml

Please test the latest appimage pinned in the release

danielcadev:
Thanks for sharing the workflow. I tested the pinned 0.9.115 AppImage on Fedora and openSUSE Tumbleweed.

Playback worked correctly on Fedora. On openSUSE, the application and media stack loaded, but an uncached P2P source remained on CONNECTING. I isolated that as a separate playback-flow issue rather than an AppImage packaging problem: the P2P confirmation path was not preserving forceP2p.

Using the same openSUSE environment, profile, torrent, and official AppImage libraries, the original payload remained on CONNECTING while the patched payload entered PREPARING STREAM and found a peer.

I opened #1084 with the focused fix and regression coverage. Thanks again for pointing me to the pinned AppImage.

PR #1073, opened 2026-07-27
feat(library): virtualize local grid, merge duplicate versions, add hero, genre filter and local Continue Watching

Jazzmedo:
## Summary

The local library became sluggish at ~145 items and couldn't represent a movie held in more than one quality. This fixes both, then builds three requested additions on top, plus one logo-language bug found along the way.

Now rebased onto `beta-branch`. One new dependency: `@tanstack/react-virtual` (see Notes).

## Performance

The local tab rendered every item at once, and every toast tick or selection toggle re-rendered all of them.

- **Virtualized** the grid with a new `VirtualGrid` component (`@tanstack/react-virtual`), threading the scroll ref down from `LibraryView`. Previously each item mounted a card carrying ~10 hooks and 5 effects through `Poster`.
- **Memoized** `OwnedCard` / `ShowGroupCard` and stabilized `cardProps`. Selection is now passed as a per-card boolean rather than the `Set`, which was replaced wholesale on every toggle and defeated memoization across the entire grid. The React Compiler isn't enabled, so this is manual.
- **Cached the store** — `read()` in `local-library.ts` re-parsed the whole library out of `localStorage` on every call, once per subscriber per mutation.
- **Cheapened hover** — `will-change`/`translate3d` on scaled posters, targeted transitions instead of `transition-all`, and no `backdrop-filter` on always-mounted card chrome.

## Duplicate versions in one entry

Two rips of the same film appeared as two posters, and only the first was ever reachable (`findLocalMovie` used `.find()`).

- Movies group by metadata id, falling back to normalized title + year, into a single poster with a version-count pill.
- Local filenames now go through the **real stream parser** rather than the scanner's regex, so version rows get source / codec / HDR / audio facets. File size is kept from the scan — `harbor_scan_folder` already returned it and TypeScript discarded it.
- Play on a multi-version entry opens a picker showing each file with the same badge stack as addon streams. It sits *after* the local-vs-stream decision, and autoplay still takes the best version rather than blocking on a dialog.
- Fixes a silent bug: the local episodes modal keyed episodes as `Map<number, LocalEntry>`, so a second file for the same episode was **invisible**, not merely unselectable.

## Genre filter (local tab)

- Genres are captured on scan from `.nfo` `<genre>` tags, falling back to TMDB. The TMDB side costs **no extra request** — that detail response was already being fetched for `imdb_id`/`runtime` and its `genres` array thrown away. This also closes a round-trip gap, since `export.ts` was already *writing* `<genre>` tags nothing read back.
- Multi-select menu in the local toolbar, ORing selections, with per-genre counts.
- Not backfilled by design, so the empty state explains why previously-scanned entries don't match.

## Continue Watching in the local collection

- A row above the local grid reusing `Row`, `ContinueCard` and the existing CW predicates (`isCwMember`, `cwSortKey`, `isCwDismissed`) — no parallel logic.
- Series collapse to one card and advance only to a next episode that **exists on disk**. Deliberately not `useCwAdvance`, which resolves episode lists over TMDB and can't handle a `local:` id.
- `ContinueCard` gains an optional `onPlayOverride` so the row plays the actual file; the default path silently falls back to the streaming picker. It is optional, so Home / Movies / Shows / Anime are unchanged.

## Featured hero on the library page

- `HeroCarousel` above the library tab bar, fed by whichever tab is open through a small context, with picks held stable for the session.
- `Slide.rank` is now optional — the "#N in … Today" trending pill doesn't apply to a personal library.
- Not shown on MAL / AniList, whose entries aren't `Meta`s with resolvable ids.

## Logo language fix

`Hero` resolved logos itself and only took the language-aware TMDB path for `tmdb:` ids. History and Trakt produce metas with **IMDb (`tt…`) ids**, so they fell through to Cinemeta and always showed the English logo regardless of the configured image language. It now uses `resolveLogo` / `peekCachedLogo`, which map `tt` ids to TMDB first and honour `tmdbImageLangs`. Since `Hero` is shared, this also corrects Home for any `tt`-id slide.

## Testing

Verified against `beta-branch`, using the same steps and tool version CI uses:

- `vp fmt --check` on all 39 changed files (vite-plus 0.2.6, matching `setup-vp@v1`) — clean
- `vp check --no-fmt` on the changed code files — 0 errors, 7 warnings, all pre-existing in untouched regions
- `tsc --noEmit -p tsconfig.json` — 15 errors, byte-identical to the `beta-branch` baseline; **0 introduced**
- `node --test tests/*.test.ts` — 79 pass / 3 fail, the same 3 that fail on clean `beta-branch` (`player-buffer-policy`, `rtx-hdr-policy`, an auto-sync subtest). 14 of the passing tests are new here.
- `pnpm install --frozen-lockfile` — passes with the updated lockfile

No Rust changes, so `cargo check` isn't applicable.

## Notes for reviewers

- **New dependency.** `beta-branch` has neither `@tanstack/react-virtual` nor a `VirtualGrid` component, so both arrive here — the dependency in its own commit (`build:`) so it can be reviewed separately from the feature. Happy to drop virtualization and keep the plain grid if you'd rather not take the dependency; the rest of the PR does not need it.
- **Formatting churn.** `App.tsx`, `library.tsx`, `local-tab.tsx` and `package.json` are already unformatted on `beta-branch`, and CI runs `vp fmt --check` over changed files, so touching them forces a reformat. `git diff -w --ignore-blank-lines` shows the real change is 2161 lines of the 2488 total; `App.tsx` in particular is ~2 real lines and the rest is formatter output.
- **`LocalTab` takes an optional `scrollRef`.** `cast-modal/library-panel.tsx` renders `<LocalTab />` with no scroll container, so that path keeps the non-virtualized grid rather than being forced to invent a scroller.
- Existing libraries are **not** migrated. New fields (`size`, `genres`) are optional and previously-scanned entries keep working; genres populate on future scans only.
- Conflicts with `beta-branch`'s refactors were resolved in favour of beta throughout — beta's `useLocalExport` hook replaced an inline export block, and beta's shift-range `selectItems` is kept and routed through the memoized card props.

## Screenshots
Screenshot-2026-07-28_01-33-05
image

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,10 +1,9 @@
 {
   "name": "harbor",
-  "private": true,
   "version": "0.9.115",
-  "type": "module",
+  "private": true,
   "license": "MIT",
-  "packageManager": "pnpm@11.9.0",
+  "type": "module",
   "scripts": {
     "dev": "vite",
     "build": "tsc -b && vite build",
@@ -21,6 +20,7 @@
   },
   "dependencies": {
     "@mediapipe/tasks-vision": "0.10.35",
+    "@tanstack/react-virtual": "^3.14.6",
     "@tauri-apps/api": "^2",
     "@tauri-apps/plugin-deep-link": "^2.4.9",
     "@tauri-apps/plugin-dialog": "^2.7.1",
@@ -56,6 +56,7 @@
     "vite": "^7.0.4",
     "vite-plugin-static-copy": "^2.0.0"
   },
+  "packageManager": "pnpm@11.9.0",
   "pnpm": {
     "onlyBuiltDependencies": [
       "esbuild"
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -11,6 +11,9 @@ importers:
       '@mediapipe/tasks-vision':
         specifier: 0.10.35
         version: 0.10.35
+      '@tanstack/react-virtual':
+        specifier: ^3.14.6
+        version: 3.14.9(react-dom@19.2.5(react@19.2.5))(react@19.2.5)
       '@tauri-apps/api':
         specifier: ^2
         version: 2.10.1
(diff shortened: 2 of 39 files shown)


harborstremio:
we will merge this, and will test this out if this needs a revert and reintegrate we will do it carefully with you since it touches CW

PR #1079, opened 2026-07-28
feat(packaging): publish Harbor to WinGet on stable releases

Jazzmedo:
## Summary

Adds Harbor to the [Windows Package Manager](https://github.com/microsoft/winget-pkgs), with a workflow that submits every stable release automatically.

```powershell
winget install HarborStremio.Harbor
winget upgrade HarborStremio.Harbor
```

Windows is currently the only supported platform without a package manager entry. Linux has `.deb`, `.rpm`, Flatpak and the AUR through the community builds repo; Windows users have to find the Releases page and download an installer by hand, then rely on the in-app updater.

Nothing existing changes except one row in the README install table. `tauri-build.yml`, `ci.yml`, `app-build.yml` and `flatpak.yml` are untouched, and the manual release process stays exactly as it is.

## What's included

| File | Purpose |
| --- | --- |
| `.github/workflows/winget.yml` | Submits a `winget-pkgs` PR on every stable release |
| `packaging/winget/*.yaml` | Bootstrap manifests for 0.9.21 (v1.10.0 schema) |
| `packaging/winget/README.md` | Maintainer setup steps and known limitations |
| `README.md` | One row in the install table |

## Package identifier

`HarborStremio.Harbor`, **not** the Flatpak's `site.harbor.Harbor`.

WinGet identifiers are `Publisher.Package`, not reverse-DNS. `site.harbor.Harbor` is schema-valid (the pattern permits 2–8 dot-separated segments) but files the package under `manifests/s/site/harbor/Harbor/` with a publisher segment reading as "site", which moderators routinely push back on. `HarborStremio.Harbor` matches the GitHub organisation and lands at `manifests/h/HarborStremio/Harbor/`. The Flatpak app ID is unchanged.

## Setup required before this can run — maintainers only

I can't do these from a fork, and the workflow stays inert until they exist:

1. **Fork `microsoft/winget-pkgs` into the `harborstremio` org.** The action pushes its branch there before opening the upstream PR. If you'd rather host the fork elsewhere, change `fork-user:` in the workflow.
2. **Add a `WINGET_TOKEN` repository secret** — a *classic* PAT with `public_repo` scope. Fine-grained tokens don't work with the fork-and-PR flow.
3. **Submit the first manifest by hand.** `winget-releaser` deliberately fails if the package isn't already in `winget-pkgs` — its first step checks the manifest directory and exits on a 404. The files in `packaging/winget/` are ready for exactly this; `komac` or `wingetcreate` will do the fork-and-PR.

After that, publishing a stable release is all that's needed.

## Design notes

**Stable only, by construction.** The trigger is `release: released`, which fires only for non-draft, **non-prerelease** publishes — so the stable-only gate needs no extra condition. This is also the only thing that *can* work today: the beta line (0.9.11x) is distributed through `https://harbor.site/updates/latest.json` and is never cut as a GitHub Release, so there's no artifact for a manifest to reference. Publishing beta to WinGet would first require beta GitHub Releases.

**Forks skip rather than fail.** The job carries `if: github.repository == 'harborstremio/harbor'`, so merging this can't turn anyone's fork CI red over a secret they don't have.

**The version is derived, not assumed.** `winget-releaser` defaults to stripping a lowercase `v` from the tag. Tag naming has drifted — `V0.9.21` has a capital V, `v.0.8.3-beta` has a stray dot, older tags carry `-beta` suffixes — so a small shell step normalises it. I ran it against all 30 existing tags; every one yields clean semver.

**A missing installer is a skip, not a failure.** If a release has no `_x64-setup.exe`, the workflow warns and exits 0.

**`Scope: user` is load-bearing.** The NSIS bundle is built `"installMode": "currentUser"` in `src-tauri/tauri.conf.json`, so Harbor installs per-user. Declaring machine scope would make `winget upgrade` misbehave.

**x64 only.** `tauri-build.yml` declares an `aarch64-pc-windows-msvc` matrix row, but no ARM64 asset has ever been attached to a release.

## Verification

| Check | Result |
| --- | --- |
| YAML parses (4 files) | pass |
| Manifests vs. winget v1.10.0 JSON schemas | pass, all three |
| Version normaliser vs. all 30 historical tags | all yield clean semver |
| `_x64-setup\.exe$` regex vs. all 30 releases | exactly one match each |
| Workflow shell step, executed locally | correct on `V0.9.21`, `v.0.8.3-beta`, `v0.8.4-beta` |
| No-asset skip path | warns, `found=false`, exit 0 |
| `vp fmt --check` on all touched files | clean |

Two things I want to be upfront about:

- **The workflow itself has not been executed end to end**, and can't be until the fork and `WINGET_TOKEN` exist. I ran its shell step locally against live releases, including the skip path, but the `winget-releaser` invocation is unexercised. The fork guard is what makes merging an untested workflow safe — it cannot run anywhere it isn't configured.
- **The `InstallerSha256` came from GitHub's published `digest` field** for the release asset rather than from re-hashing the 280 MB installer locally. That digest is computed by GitHub over the stored file, but it's worth a `winget hash` before the bootstrap PR goes up.

`actionlint` and `shellcheck` weren't available in my environment, so the workflow is not statically linted.

## Out of scope, but noticed

- **Windows installers are unsigned.** No signing secret exists in any workflow. WinGet accepts unsigned installers, but expect SmartScreen warnings and the occasional AV false positive during validation.
- **The in-app updater doesn't know about WinGet.** A WinGet-installed user still gets in-app update prompts. Linux already defers to the package manager via `isLinuxDesktop()`; an equivalent WinGet check would be a sensible follow-up.
- **`matrix.platform.bundle: msi` in `tauri-build.yml` is inert** — it's declared but never passed to any step, so Windows actually emits NSIS `.exe` per `tauri.conf.json`. Harmless today since the upload step globs both, but the `*.msi` glob is dead. Left alone deliberately; happy to file it separately.

## Testing after setup

Once the manifest is live upstream:

```powershell
winget search Harbor
winget install HarborStremio.Harbor
winget list HarborStremio.Harbor
```

Then publish a stable release and confirm the workflow opens a PR against `microsoft/winget-pkgs` with the new version.

### DIFF
--- a/.github/workflows/winget.yml
+++ b/.github/workflows/winget.yml
@@ -1,0 +1,72 @@
+name: WinGet
+
+on:
+  release:
+    types:
+      - released
+  workflow_dispatch:
+    inputs:
+      tag:
+        description: Release tag to publish to WinGet
+        required: true
+
+permissions:
+  contents: read
+
+concurrency:
+  group: winget-${{ github.workflow }}-${{ github.event.release.tag_name || inputs.tag }}
+  cancel-in-progress: false
+
+jobs:
+  publish:
+    name: Publish
+    # Only the upstream repository has the winget-pkgs fork and the WINGET_TOKEN
+    # secret, so forks skip instead of failing on a missing credential.
+    if: github.repository == 'harborstremio/harbor'
+    runs-on: ubuntu-24.04
+    timeout-minutes: 20
+    steps:
+      - name: Resolve release
+        id: release
+        env:
+          GH_TOKEN: ${{ github.token }}
+          TAG: ${{ github.event.release.tag_name || inputs.tag }}
+        run: |
+          set -euo pipefail
+
+          # Release tags are not consistently formatted (V0.9.21, v0.9.20,
+          # v.0.8.3-beta), so derive the WinGet PackageVersion rather than
+          # relying on the action's default "strip a leading v".
+          version="${TAG#[Vv]}"
+          version="${version#.}"
+          version="${version%%-*}"
+          if [ -z "$version" ]; then
+            echo "::error::Could not derive a version from tag '$TAG'."
+            exit 1
+          fi
+
+          # A release published without a Windows build is not an error; there is
+          # simply nothing to submit.
+          installer="$(gh release view "$TAG" --repo "$GITHUB_REPOSITORY" \
+            --json assets --jq '[.assets[].name | select(test("_x64-setup\\.exe$"))] | first // ""')"
+          if [ -z "$installer" ]; then
+            echo "::warning::No _x64-setup.exe asset on $TAG; skipping WinGet submission."
+            echo "found=false" >> "$GITHUB_OUTPUT"
+            exit 0
+          fi
+
+          echo "Submitting $installer as version $version."
+          echo "found=true" >> "$GITHUB_OUTPUT"
+          echo "tag=$TAG" >> "$GITHUB_OUTPUT"
+          echo "version=$version" >> "$GITHUB_OUTPUT"
+
+      - name: Submit to winget-pkgs
+        if: steps.release.outputs.found == 'true'
+        uses: vedantmgoyal9/winget-releaser@v2
+        with:
+          identifier: HarborStremio.Harbor
+          version: ${{ steps.release.outputs.version }}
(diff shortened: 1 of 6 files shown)


harborstremio:
works on our end thank you :D

PR #1081, opened 2026-07-28
fix(security): an encrypted secrets vault for debrid keys

GaintCatPerson:
### Summary

This PR moves all debrid API keys (Real-Debrid, AllDeebrid, Premiumize, Torbox, DebridLink) out of the plaintext `settings.json` file and into an encrypted secrets vault.

Today, a user's debrid credentials live unencrypted inside `settings.json` (alongside every other preference). Anyone or anything that can read that one file — malware, a malicious or compromised backup/sync client, a config-share tool, a screenshot/diagnostics exporter, or simply another local user on a shared machine — obtains full, usable access to the user's paid debrid accounts. Debrid keys are effectively account credentials: with them, one can stream, download, and incur costs on the user's subscription.

We close that gap by encrypting the secrets at rest.

### Why this is necessary (security risk)

- **Single-file exposure = full credential theft.** `settings.json` is flat, unencrypted JSON. A single compromised read yields every debrid key in cleartext.
- **The threat is realistic, not theoretical.** Harbor already ships a "Bug Report / diagnostics" feature that enumerates configured services. Before this change it even counted debrid keys (`debridCount`) and exposed addon configuration. Leaking the diagnostic payload (intentionally or via a pastebin) would hand over working keys. We now read those counts from the vault and never surface key material.
- **Cloud/backup sync is the norm.** Many users sync `%APPDATA%` or the config directory across machines (OneDrive, Dropbox, rsync, Syncthing). An unencrypted `settings.json` in a sync folder is a credential in the cloud.
- **Defense-in-depth for the project's future health.** As Harbor grows (more services, more platforms, TV/headless deployments), the "just put secrets next to prefs" pattern does not scale and keeps accruing risk. Establishing a proper vault now prevents a larger, harder migration later.

### What this PR does

1. **Removes debrid keys from the `Settings` type** (`types.ts`, `defaults.ts`). They can no longer be serialized into `settings.json`.
2. **Adds an encrypted vault in Rust** (`settings_store.rs`):
   - Each secret is encrypted with **AES-256-GCM** under a per-install random key (KEK) stored at `secrets.kek`.
   - The ciphertext is written to `secrets.vault.json` — a self-contained, OS-independent envelope that is *always* the source of truth.
   - When an OS credential store is available (Windows Credential Manager, macOS Keychain, Linux Secret Service), each key is **also mirrored per-key** into the keyring for tighter OS integration. This is best-effort only; the envelope remains authoritative.
   - **Headless / no-keyring environments** (some Linux setups, Flatpak without a portal) transparently fall back to the envelope alone — no breakage.
3. **Adds a small reactive TypeScript layer** (`secret-store.ts`) so the UI can read/write debrid secrets and stay in sync (`useDebridSecrets`, `getSecret`, `setSecret`), with a localStorage bridge for the web/no-backend build so behavior stays identical.
4. **Rewires every read site** (addon construction, auto-download, streaming panel, settings nav counts, basics panel, bug-report diagnostics) to read from the vault.
5. **Migrates existing installs** (`load.ts`): on first load, any legacy plaintext debrid key in `settings.json` is moved into the vault and deleted from the parsed settings, so it is never re-persisted. The migration is **idempotent** and safe to run repeatedly.

### Note on keyring size limits

A design note worth flagging for reviewers: the OS keyring is used only as a best-effort per-key mirror of each secret, never as the source of truth. This is deliberate — some backends (notably Windows Credential Manager) cap a single credential blob near 512 bytes, which a full secret map can exceed. Keeping the AES-256-GCM envelope (`secrets.vault.json`) authoritative and writing the keyring one small entry per key sidesteps that cap entirely and avoids any risk of a stale or truncated blob being served on read.

### Trade-offs / risks considered

- **Key material location.** The KEK lives in `secrets.kek` on the same machine as the vault. This is defense-in-depth, not perfect isolation: anyone with full disk + keyring access can still decrypt. That is the same trust boundary as the OS keyring itself, and it is a strict improvement over plaintext-in-`settings.json`. Full OS-user-isolation would require a TPM/secure-enclave path, which is out of scope for this change and a natural follow-up.
- **No new files / minimal surface.** All Rust logic folds into the existing `settings_store.rs`; all TS logic folds into the existing `secret-store.ts`. The only added files are the two vault artifacts on disk (`secrets.kek`, `secrets.vault.json`), which are runtime data, not source.
- **Flatpak.** Adds an optional `gnome-keyring` module so the keyring backend has a Secret Service to talk to; it is a soft dependency — the envelope fallback covers its absence.

### Verification

- `pnpm run build` (`tsc -b && vite build`) — passes, `BUILD_EXIT=0`.
- `cargo check` (via the standard Tauri build path) — passes with only the two pre-existing `vapoursynth` warnings, no new warnings.
- `tauri.conf.json` / `tauri.windows.conf.json` left unchanged.
- Scope: 21 files, debrid keys confirmed absent from `settings.json` persistence and from the bug-report diagnostic payload.


AdityaHebballe:
- Debrid credentials are no longer profile-scoped. loadStoredSettings() migrates each profile’s rdKey/etc. into the same global names (harbor.debrid.rdKey, etc.). When switching from profile A to B, B overwrites the global vault values; switching back to A then uses B’s credentials after A’s plaintext settings have been persisted without keys. This both loses A’s keys and can expose one profile’s paid account to another. The vault keys need a profile namespace, or debrid credentials need an explicitly global product decision and migration.

- Existing secrets.json data is discarded on upgrade. The old native secret store reads secrets.json; this PR changes reads to only secrets.vault.json and contains no migration for the old file. Existing Simkl, Trakt, MAL, and AniList secrets therefore disappear from the app after upgrading, while the old plaintext file remains on disk. Migrate and remove/secure the legacy file atomically before switching formats.

- The Linux “encryption at rest” does not protect against the stated local-user threat. The KEK is written beside the encrypted vault using std::fs::write, and the app-data directory is created with default permissions. With the usual umask this results in readable files/directories (commonly 0644 files / 0755 directory). A local user who can read secrets.kek can decrypt secrets.vault.json, so this does not address the PR’s shared-machine claim. Create the directory as 0700 and KEK/vault/temp files as 0600 on Unix (and handle Windows ACLs appropriately), or store the KEK in the OS credential store.

After these are clarified I will test this PR. Thank you for your contribution

GaintCatPerson:
Closing — stale, author cleanup on return. Will re-open fresh if still needed.

PR #1084, opened 2026-07-28
fix(playback): force local engine after P2P confirmation

danielcadev:
## Summary

Forces the local torrent engine when the user confirms playback from the peer-to-peer confirmation dialog.

Adds a focused regression test covering both paths that open this dialog.

## Why

The confirmation dialog is shown for an engine-eligible uncached torrent, but one of its call sites did not preserve the `forceP2p` intent. After confirming, Harbor could run the regular resolution path again and remain on the `CONNECTING` screen instead of starting local peer discovery.

This change passes `forceP2p: true` through the existing confirmation state, matching the fallback path already present in the same handler.

## Verification

- `vp fmt --check` for the changed files: passed.
- `vp check --no-fmt` for the changed files: passed.
- `pnpm exec tsc -b --pretty false`: passed.
- `node --test tests/p2p-confirmation.test.ts`: passed.
- Manual A/B test using the same openSUSE Tumbleweed container, profile configuration, torrent, and official AppImage libraries:
  - Official payload remained on `CONNECTING`.
  - Patched payload entered `PREPARING STREAM` and found a peer.
- The Linux release binary built successfully. The final AppImage packaging stage still encounters the existing `linuxdeploy` packaging failure.

## Platform Impact

Verified on Linux using an openSUSE Tumbleweed container. The change is in the shared TypeScript playback flow and does not introduce platform-specific behavior.

## Checklist

- [x] I ran TypeScript typechecking after TypeScript changes.
- [x] No Rust files changed, so Cargo checks were not applicable.
- [x] I tested the affected behavior on Linux.

### DIFF
--- a/src/views/play-picker/use-pick-handler.ts
+++ b/src/views/play-picker/use-pick-handler.ts
@@ -82,7 +82,9 @@ export function usePickHandler({
 }) {
   const [queuedHash, setQueuedHash] = useState<string | null>(null);
   const [debridDown, setDebridDown] = useState(false);
-  const [p2pConfirm, setP2pConfirm] = useState<{ stream: ScoredStream; forceP2p?: boolean } | null>(null);
+  const [p2pConfirm, setP2pConfirm] = useState<{ stream: ScoredStream; forceP2p?: boolean } | null>(
+    null,
+  );
   const debridFailStreakRef = useRef(0);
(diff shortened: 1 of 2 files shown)


harborstremio:
So sorry for the delays busy week thank you so much for your effort :D ! @danielcadev

PR #1092, opened 2026-07-29
fix(subtitles): keep the second subtitle visible under HDR passthrough

Jazzmedo:
## Summary

On a Windows HDR passthrough surface nothing can be composited over the video. That is why `suppressHtmlSubs` folds `hdrNativeSurface` in (`use-player-media.ts:161`) and the primary subtitle is handed to mpv through `setSubVisible(subNativeRender)`.

The second subtitle had no equivalent. `secondary-sub-visibility` is set to `"no"` at startup (`src-tauri/src/mpv.rs:717`) and never turned back on, so on that one surface the second subtitle is simply invisible, with nothing indicating why. Every other configuration is unaffected.

## Change

Mirrors the arrangement the primary already has.

`applySecondarySubNative` sits next to `applySubStyle` in `sub-style.ts` — the module that already owns mpv's subtitle properties — and drives `secondary-sub-visibility` plus `secondary-sub-pos`. `useSecondarySub` makes the call, so the whole secondary lifecycle stays in the one hook that owns it.

`secondary-sub-pos` is measured from the top, matching how `sub-pos` positions the primary. **Top of the screen** maps to `0`; **Above the main line** offsets from the configured margin by a fixed slice of the frame, because mpv exposes no line-height metric to be exact with.

The overlay is deliberately left untouched. It cannot draw on that surface, and the HDR chrome window renders `ShellLayer` only — not `StageOverlays` — so there is no second renderer that could double up. An earlier draft threaded a flag through `player.tsx` to blank `secondarySubText`; I dropped it once that turned out to guard against nothing, which also keeps `player.tsx` out of the diff entirely.

## Commits

- `fix(subtitles):` the change itself — 3 files, +35/−0.
- `style:` formatting. `vp fmt --check` in the Frontend job runs over changed files, and `sub-style.ts` and `use-player-media.ts` were both already unformatted on `beta-branch`, so touching them at all turns CI red. Split out so the two-line change in `use-player-media.ts` stays visible.

## Testing

`tsc -b` clean. `vp fmt --check` clean on all three files.

**I could not verify this on hardware** — it needs a Windows machine with an HDR display and a PQ/HLG source, which I don't have. The change mirrors the primary's existing path and the `hdrNativeSurface` guard means nothing else can reach it, but the on-screen result on that surface is unconfirmed. Happy to adjust the `secondary-sub-pos` offset if it sits wrong against the primary line.

### DIFF
--- a/src/lib/player/sub-style.ts
+++ b/src/lib/player/sub-style.ts
@@ -30,6 +30,25 @@ function mpvFontFor(id: string): string {
   }
 }
 
+/** Hand the second subtitle over to mpv's own renderer, for the one surface our
+ *  overlay cannot draw on (HDR passthrough). `secondary-sub-pos` is measured from
+ *  the top, matching how `sub-pos` positions the primary. */
+export async function applySecondarySubNative(
+  on: boolean,
+  placement: Settings["subSecondaryPlacement"],
+  marginY: number,
+): Promise<void> {
+  await invoke("mpv_set_property", {
+    name: "secondary-sub-visibility",
+    value: on,
+  }).catch(() => {});
+  if (!on) return;
+  // "bottom" means "just above the primary line". mpv exposes no line-height
+  // metric, so offset by a fixed slice of the frame rather than guessing exactly.
+  const pos = placement === "top" ? 0 : clamp(100 - (Number(marginY) || 0) - 8, 0, 100);
+  await invoke("mpv_set_property", { name: "secondary-sub-pos", value: pos }).catch(() => {});
+}
+
 export type SubRenderContext = {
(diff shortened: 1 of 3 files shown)


3-pr:
Hey @Jazzmedo Thanks for your awesome contributions! Just wondering if you have Discord so we could chat about some work on Harbor 👀

Jazzmedo:
> Hey @Jazzmedo Thanks for your awesome contributions! Just wondering if you have Discord so we could chat about some work on Harbor 👀

I'm already following you on X and I'm interested, Here is my email 

mohamedmelawany@hotmail.com

PR #1097, opened 2026-07-30
docs: mark Linux builds as official

AdityaHebballe:
## Official Linux builds

Automatic updates are now enabled for the `.deb`, `.rpm`, and AppImage packages, so these can be considered official Harbor Linux builds.

Flatpak support is underway and will be added to the repository soon.

This updates the README to use the official packaging repository and adds the Cloudsmith OSS-hosting attribution.

### DIFF
--- a/README.md
+++ b/README.md
@@ -42,7 +42,7 @@ Offering things like a native player, a stream ranking engine, Wikidata, watch p
 > <br/>
 
 > [!NOTE]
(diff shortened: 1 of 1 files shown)


JJDizz1L (approved):
Over the last few days both on AMD and NVIDIA, Harbor has been solid. Will test the AppImage later. But as long as yay/paru works for Arch, that covers 98% of Arch users.

AdityaHebballe:
Thank you @JJDizz1L. A lot of effort has gone into making Appimage work, it currently uses sharun which is not even merged into tauri and is a draft PR.

For this reason alone I call it experimental but apart from that it appears stable. If you could test AppImage as well it would be a great help. Thank you!

.deb and .rpm are very stable and can be updated using their respective package manager thanks to cloudsmith's repo.

PR #1195, opened 2026-08-08
manga: add manual library refresh button

AbdullahAljehani:
## Summary

Add a manual refresh button to the Manga library view. Clicking it clears the manga cache and re-fetches from the active source (e.g. a connected Suwayomi server) immediately.

## Why

Manga library data is cached for up to 30 minutes before a background refresh is even attempted, and up to 6 hours in the worst case (see POPULAR_DISK / mangaChapters TTLs in src/lib/manga/api.ts). There is currently no manual refresh action anywhere in the Manga UI. This means new chapters on a connected Suwayomi server (e.g. a weekly release) don't show up in Harbor until the entire app is restarted, which clears the in-memory cache. This button lets users force a fresh fetch on demand without restarting.

## Verification

- Ran `pnpm install` and `npx vite build` — build completes with no errors, no TypeScript issues in the changed file.
- Reviewed `clearMangaCache()` in src/lib/manga/api.ts to confirm it clears both the in-memory cache and the disk-backed cache, so the fix covers both the library list and per-manga chapter lists.
- Have not yet run this live in `tauri dev` against a real Suwayomi server end to end — happy to do so if a maintainer wants that confirmed before merge, or if someone else can verify.

## Platform Impact

None expected. Pure frontend (React/TypeScript) change, no platform-specific code touched.

## UI Changes

Added a small circular refresh icon button next to the source/tag filters on the Manga library screen (top toolbar). Spins while refreshing, disabled until the fetch completes or fails. No screenshot yet — can add one if requested.

## Checklist

### DIFF
--- a/src/views/manga/manga-browse.tsx
+++ b/src/views/manga/manga-browse.tsx
@@ -1,8 +1,9 @@
-import { Loader2, Search, Star } from "lucide-react";
+import { Loader2, RefreshCw, Search, Star } from "lucide-react";
 import { useCallback, useEffect, useMemo, useRef, useState } from "react";
 import { useT } from "@/lib/i18n";
 import { Poster } from "@/components/poster";
 import {
+  clearMangaCache,
   MANGA_PAGE,
   popularManga,
   popularMangaStream,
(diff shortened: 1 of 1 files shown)


Auramist:
hey @AbdullahAljehani , great work and this was indeed needed. thanks a lot. I just wanted to ask if it's possible to make the feature work whenever the manga navigation is clicked. I think this way it will be more convenient instead of clicking the refresh button each time.
 
 It would be a quality-of-life addition if that's possible of course.

PR #1220, opened 2026-08-13
Fix: controller input firing while window unfocused

Thunderhawkk:
## Summary

Harbor reads controller input from two sources: the native gilrs source (Rust) and, on Windows, a Web Gamepad API fallback for pads not handled natively (non-XInput devices such as DualShock/DualSense, Switch Pro, 8BitDo, etc.).

The native source was already focus-gated via gamepad_set_background_input / WindowEvent::Focused, but the web source polled navigator.getGamepads() in a requestAnimationFrame loop and dispatched button/axis events with no focus check at all. While Harbor was unfocused, the WebView kept polling and drove navigation in the background window.

This change applies the same focus policy the native side already enforces to the web source:

src/lib/gamepad/web-source.ts — added an optional inputAllowed predicate to WebGamepadHandlers (defaults to allow-all). Button/axis dispatch is gated on it. Pad state is still read every frame so the internal pressed/axis maps stay in sync (no spurious or stale events on refocus); onPads remains ungated so the connected-device list always updates, matching the native Connected/Disconnected events.
src/lib/gamepad/use-gamepad.ts — passes inputAllowed: () => document.hasFocus() || backgroundRef.current, reusing the same document.hasFocus() signal hotkeys.ts already relies on and honoring the existing "controller background input" setting.
## Why

Controller input leaked into the unfocused Harbor window while the user was driving the same controller in another app. The web gamepad source was the only input path that ignored both window focus and the controllerBackgroundInput setting. Fixing it at the source keeps behavior identical to the already-correct native path instead of adding a frontend-side workaround.

## Verification

tsc --noEmit -p tsconfig.json — passes (exit 0).

## UI Changes

None.

## Checklist

- [x] I ran vp check for the changed files.
- [x] I ran vp run typecheck after TypeScript changes.
- [ ] I added or updated tests for behavior changes. (no TS test infra in repo; change is a small gate)

### DIFF
--- a/src/lib/gamepad/use-gamepad.ts
+++ b/src/lib/gamepad/use-gamepad.ts
@@ -67,6 +67,9 @@ export function useGamepad(): void {
   const playerRef = useRef(!!player);
   playerRef.current = !!player;
 
+  const backgroundRef = useRef(backgroundInput);
+  backgroundRef.current = backgroundInput;
+
   useEffect(() => {
     if (!isTauri || !enabled) return;
 
@@ -174,6 +177,7 @@ export function useGamepad(): void {
     });
 
     const stopWebSource = startWebGamepadSource({
+      inputAllowed: () => document.hasFocus() || backgroundRef.current,
       onButton: (button, isPressed) => {
         setLiveButton(button, isPressed);
         onButton(button, isPressed);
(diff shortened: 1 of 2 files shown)


Talal1011:
added to PR #1280

PR #1229, opened 2026-08-16
feat(subtitles): add native source FPS control

OwaisByte:
## Summary

- Add a manual **Subtitle source FPS** control beside the existing subtitle sync control.
- Use a dedicated lightweight subtitle-timing icon instead of a generic gauge, matching the control's actual purpose without adding an asset or dependency.
- Apply the selected rate through mpv's native `sub-fps` property without rewriting subtitle files.
- Read Video FPS from `estimated-vf-fps`, with `container-fps` as the same fallback used by Playback Stats.
- Pass the active player engine through both transport renderers, gate the control on an active, non-idle mpv session in the main Tauri window, and hide it again on mpv end/shutdown events.
- Reset an active correction when media or the primary subtitle track changes, secondary subtitles become active, the player shuts down, or Auto Sync starts or applies a result.
- Route HDR-overlay subtitle selection and subtitle addition back through the main `PlayerBridge`, with a media-identity guard so delayed actions from an old overlay cannot affect current playback.

## Why

A constant subtitle delay cannot fix timing that drifts progressively because the subtitle and video were authored at different frame rates. This control lets the user provide the subtitle's known source frame rate while Harbor keeps the video's frame rate, playback rate, and existing subtitle delay unchanged.

`No correction (default)` maps to mpv's `sub-fps=0`. It replaces the misleading idea of an automatic source-FPS detector: Harbor does not infer a subtitle source frame rate from timestamp-only subtitle data.

## Verification

- Fresh reviewer checkout: `pnpm install --frozen-lockfile` and the repository's `pnpm setup` completed successfully.
- `cargo check --manifest-path src-tauri/Cargo.toml`: passed after the official setup downloaded the bundled mpv runtime; two existing dead-code warnings remain in untouched Rust code.
- `vp fmt --check <all 23 changed files>`: passed.
- `vp check --no-fmt <all 23 changed files>`: passed with no warnings or lint errors.
- `pnpm exec tsc -b --pretty false`: passed.
- `node --test --experimental-strip-types tests/subtitle-fps.test.ts tests/autosync-switch-retrigger.test.ts`: 28/28 passed.
- `pnpm build`: passed; 4,706 modules transformed.
- Full repository suite: 296/302 passed. The exact base commit passes 271/277; the same six unrelated failures reproduce in both fresh checkouts, so this change adds 25 passing tests and no new failure.

Manual checks before merge:

- [ ] Select a known 25 FPS text subtitle over a 23.976 FPS video and confirm progressive drift is corrected during normal playback.
- [x] Apply a 25 FPS source value over 23.976 FPS Windows libmpv playback and confirm it is accepted without an error.
- [x] Switch the primary subtitle track and confirm the control returns to `No correction`.
- [ ] Switch media and confirm the previous source FPS does not carry over.
- [ ] Start Auto Sync and confirm the manual FPS correction is cleared first.
- [ ] Confirm bitmap and secondary-subtitle states show a disabled explanation instead of applying a correction.

## Platform Impact

The control is limited to desktop libmpv playback in the main player UI. Both the standard and Stremio transport renderers pass their actual engine to the subtitle menu. The control then verifies that mpv is running a loaded file through `idle-active` and follows `file-loaded`, `end-file`, and `shutdown` events so it does not remain available after playback stops. It is hidden in HTML5-only playback and separate modal player shells. Subtitle actions from Harbor's HDR overlay are forwarded to the main bridge with the originating media key, so delayed track or subtitle actions cannot cross a media boundary.

Text tracks classified as SRT/SubRip, ASS, SSA, WebVTT, or MOV_TEXT are supported by the code path. Embedded text tracks use the same path but still need runtime fixture coverage. Bitmap subtitles such as PGS, VobSub/DVD, DVB, and XSUB are intentionally unsupported. Secondary subtitles are intentionally unsupported because mpv's `sub-fps` property is global.

There are no Rust changes, new commands, dependencies, subtitle-file writes, playback-rate changes, or subtitle-delay changes.

## UI Changes

The subtitle menu header gains a compact subtitle-timing button beside Sync. Its custom vector icon combines subtitle lines and a clock, while inheriting the existing button's color and states. The panel shows Video FPS, a `No correction` default, standard source-FPS presets, and a validated custom value from 1 through 240.

**Header trigger with a selected text subtitle**

Subtitle FPS trigger with a selected text subtitle

**Default timing**

Subtitle FPS panel showing No correction and 23.976 video FPS

**25 FPS subtitle source over 23.976 FPS video**

Subtitle FPS panel applying a 25 FPS subtitle source over 23.976 FPS video

## Checklist

- [ ] I ran `vp run typecheck` after TypeScript changes. The current beta branch does not define this Vite+ task; the direct `pnpm exec tsc -b --pretty false` equivalent passed.
- [x] I ran the relevant Cargo checks after Rust changes. Not applicable: this pull request has no Rust changes.
- [x] I tested affected platforms when the change is platform-specific. Verified on Windows libmpv; macOS and Linux remain untested.

### DIFF
--- a/src/components/player/autosync/autosync-store.ts
+++ b/src/components/player/autosync/autosync-store.ts
@@ -1,36 +1,12 @@
 import { useSyncExternalStore } from "react";
 import type { AutoSyncHandle } from "@/views/player/hooks/use-auto-sync";
-import { resetMpvSubtitleFpsForTransition } from "@/lib/player/mpv-properties";
 
 let current: AutoSyncHandle | null = null;
 const listeners = new Set<() => void>();
 
 export function publishAutoSync(handle: AutoSyncHandle | null): void {
-  const resetSubtitleFps = () =>
-    resetMpvSubtitleFpsForTransition((error) =>
-      console.warn("[auto-sync] subtitle FPS reset failed", error),
-    );
-  const afterSubtitleFpsReset = (action: () => void) => () => {
-    void resetSubtitleFps()
-      .then(action)
-      .catch(() => {});
-  };
-  current = handle
-    ? {
-        ...handle,
-        run: afterSubtitleFpsReset(handle.run),
-        retry: afterSubtitleFpsReset(handle.retry),
-        applyOffer: afterSubtitleFpsReset(handle.applyOffer),
-      }
-    : null;
-  if (
-    handle?.status === "analyzing" ||
-    handle?.status === "synced" ||
-    handle?.status === "best-effort" ||
-    handle?.status === "offer"
-  ) {
-    void resetSubtitleFps().catch(handle.stop);
-  }
+  if (current === handle) return;
+  current = handle;
   listeners.forEach((l) => l());
 }
 
--- a/src/components/player/autosync/autosync-store.ts
+++ b/src/components/player/autosync/autosync-store.ts
@@ -1,12 +1,36 @@
 import { useSyncExternalStore } from "react";
 import type { AutoSyncHandle } from "@/views/player/hooks/use-auto-sync";
+import { resetMpvSubtitleFpsForTransition } from "@/lib/player/mpv-properties";
 
 let current: AutoSyncHandle | null = null;
 const listeners = new Set<() => void>();
 
 export function publishAutoSync(handle: AutoSyncHandle | null): void {
-  if (current === handle) return;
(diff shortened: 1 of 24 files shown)


Talal1011:
added to PR #1280

PR #1236, opened 2026-08-17
fix(profiles): per-profile data isolation and addon refresh on profile switch

Thunderhawkk:
## Summary

Scopes all per-profile data (watched state, continue-watching progress, library sync/recovery, session and addon keying) under the active profile, and fixes stale addons/catalogs after switching profiles.

Previously, profile-scoped storage existed but UI refetches for addons and home catalogs were keyed only on the Stremio session (authKey). When profiles share the same session state (no account signed in, or the same account on every profile), authKey is identical across profiles, so switching profiles never triggered a refetch — leaving the previous profile's addons and addon-added catalogs visible until a manual refresh.

The fix listens for the existing harbor:active-profile-changed event (dispatched by selectProfile after the profile's storage is written) and forces the addons store and home rows to refetch, so each profile sees exactly its own addons and catalogs immediately on switch.

The work is split into four commits: per-profile session/addon keying, watched state + progress scoping, library sync/recovery scoping, and the addon refresh fix.

## Why

Harbor supports multiple profiles (primary, shared, separate) with isolated data. Two gaps remained:
1. Data isolation — watched flags, progress, watchlist, and sync state were not consistently scoped per profile, allowing state to leak between profiles.

2. Stale addons/catalogs — the reported bug: switching from a primary/shared profile to a separate profile (or vice versa) left the previous profile's addons and addon-added catalogs visible until refresh. Root cause: refetches keyed on authKey only, which is identical across profiles that share session state.

This approach fits Harbor's existing patterns — the fix reuses the profile-switch event that other stores already invalidate on, with no new abstractions or platform-specific code.

## Verification

- npx tsc -b --pretty false — passed (clean)
- pnpm run build — passed
- Manual scenarios:
- With no account signed in, create one separate profile; install an addon while on it
- Switch separate → primary/shared and back; verify addons and addon-added catalogs update immediately without refreshing
- Verify watched flags, continue-watching progress, and watchlist stay isolated per profile
- Verify the reverse direction (primary/shared → separate shows the separate profile's state)

## Platform Impact

None

## Checklist

### DIFF
--- a/src/components/continue-card.tsx
+++ b/src/components/continue-card.tsx
@@ -16,7 +16,7 @@ import {
 } from "@/lib/stremio";
 import { useHasNewEpisode } from "@/lib/new-episodes";
 import { Tooltip } from "@/views/detail/tooltip";
-import { useProfiles } from "@/lib/profiles";
+import { useProfiles, sharesStremioStorage } from "@/lib/profiles";
 import { useSettings } from "@/lib/settings";
 import { useView, type PlayEpisode } from "@/lib/view";
 import { getWatchedBy } from "@/lib/watched-by";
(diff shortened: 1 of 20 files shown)


Talal1011:
added to PR #1280

PR #1237, opened 2026-08-17
feat: add fullscreen in Live TV, chrome-hiding, and per-tile resizing to multiview

O1Abdulrahman:
### Multiview layout controls: fullscreen, chrome hiding, and resizable tiles

**Summary**

Improves the Multiview screen-split experience under Live TV with full control over how the split grid is displayed and sized.

**Changes**

- Fullscreen split view: New button (and Escape to exit) toggles the whole split grid to fullscreen.

- Hide top bar: The collapse button hides the app nav bar and shows only the window controls (minimize/maximize/close) while collapsed or in fullscreen.

- Resizable tiles: Drag dividers in every layout — side-by-side, stacked, triple, and quad. Each split (including left/right columns in the quad layout) is independent, so resizing one tile never moves the others.

- Align rows in quad layout: Double-click the center handle to snap both row dividers to the same position and link them together (handle highlights while linked). Dragging any row divider unlinks it so each side can be resized independently. Double-click again to re-align.

- Persisted layout: Split positions are saved per layout in localStorage and restored on next launch.

Where
Accessible from Live TV → Multiview, the split grid layout bar.

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1007,23 +1007,35 @@ function Shell({ onReady }: { onReady?: () => void }) {
       {!settingsTop && !playerActive && !liveTop && !pickerTop && layout === "stremio" && (
         <StremioRail />
       )}
-      {!settingsTop && !playerActive && !pickerTop && layout === "topdock" && <TopDock />}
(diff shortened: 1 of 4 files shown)


Talal1011:
added to PR #1280

PR #1240, opened 2026-08-17
fix(settings): scope background images per profile

uddx7:
## Summary

Fix profile-scoped background image storage. Previously, all profiles shared a single global IndexedDB key (`bg`), causing wallpaper uploads in one profile to overwrite backgrounds for all other profiles.

This change scopes background image storage by profile ID (`bg_<profileId>`), ensuring each profile maintains its own independent wallpaper.

## Why

**Problem:** When users with multiple profiles uploaded wallpapers, each upload overwrote the previous one because all profiles wrote to the same storage key. This caused:
- Loss of custom wallpapers when switching profiles
- Last uploaded wallpaper appearing on all profiles
- Poor experience for multi-profile households

**Solution:** Scope background storage by profile ID, matching how other theme settings (dim level, preset) are already stored. Each profile gets isolated storage with automatic legacy migration.

### Automated Tests
```bash
node --test tests/profile-background-persistence.test.ts  # 4/4 passing
pnpm run typecheck                                         # 0 errors
pnpm run build                                            # success

### DIFF
--- a/src/lib/backup.ts
+++ b/src/lib/backup.ts
@@ -1,5 +1,6 @@
 import { downloadText } from "@/lib/download-text";
 import { loadBgImage, saveBgImage } from "@/lib/theme-storage";
+import { readAllProfilesIdentity } from "@/lib/profiles";
 
 declare const __APP_VERSION__: string;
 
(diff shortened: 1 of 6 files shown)


Talal1011:
added to PR #1280

PR #1243, opened 2026-08-18
feat: persist addon origin in saved items and fall back to alternate stream types

Thunderhawkk:
## Summary

Saving an item to the watchlist, favorites, or a custom list only persisted id, type, name, and poster. The type was normalized to movie/series, which can drop the type an addon's catalog actually declared for non-standard id schemes. On replay, the pipeline queried /stream/<movie|series>/<id>.json, and addons that only serve that id under a different declared type returned an empty result — so saved items from those catalogs showed zero streams. The saved entries also never recorded their source addon (addonOrigin), so the pipeline could not route the query back to the origin addon.

This change does two things:

Persist source metadata — addonOrigin and embedded videos are now carried through watchlist, favorites, and custom-list entries (types, read/write, hydration, and every toggle call site), so a saved item remembers which addon served it.
Recover from collapsed types — fetchAddonStreams now retries the addon's other manifest-declared stream types when the primary query returns no streams. This runs on the forced origin-addon path and on the normal id-matching path for non-standard id schemes (where stored movie/series is unreliable). Standard schemes (tt, tmdb:, kitsu:, ...) are untouched, so no extra requests for normal content.

## Why

Old entries saved before this fix lack addonOrigin, so the origin-addon route alone was not enough. The fallback must also work through normal id-matching for non-standard id schemes — otherwise re-adding every saved item would be required. The approach is fully manifest-driven: it reads the addon's declared stream types and id prefixes rather than special-casing any content category.

## Verification

- tsc -b --pretty false — passes (exit 0).

- Reproduced against a real addon whose catalog declares additional stream types: the type the addon actually serves the id  under returns a stream list, while the collapsed movie type returns {"streams":[]} — confirming the type-collapse root cause.

- Confirmed from a fresh backup that the failing saved entries lack addonOrigin (saved pre-fix), while the one re-added item that worked carries it.

- Console trace before fix: [addons] querying 2: AIOStreams, AIOStreams then 0 streams twice. After this change the alternate declared types are retried, so a non-empty result is expected

## Platform Impact

None

## UI Changes

None

## Checklist

### DIFF
--- a/src/components/context-menu.tsx
+++ b/src/components/context-menu.tsx
@@ -193,6 +193,8 @@ export function ContextMenu() {
         name: meta.name,
         poster: meta.poster,
         imdbId: targetImdb,
+        addonOrigin: meta.addonOrigin,
+        videos: meta.videos,
       });
       close();
     };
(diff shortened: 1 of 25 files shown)


Talal1011:
added to PR #1280

PR #1244, opened 2026-08-18
feat(i18n): add French translation

Aerya:
## Summary

- add French (`fr`) to the supported Harbor UI languages
- add a complete French translation catalog
- preserve placeholders, technical names, and the existing English fallback behavior


Aerya:
Bump.
@harborstremio ?

PR #1246, opened 2026-08-18
fix(discover): apply configured region to movie discovery queries

Aerya:
## Summary

The configured **Region & language** value already participates in Discover cache scoping, but `settings.region` is currently added to movie discovery queries only when the optional locale-bias feature is enabled and its language conditions match.

That couples two independent settings.

This change applies the configured region to TMDB movie discovery queries independently of `feedLocaleBias`, while leaving the language/ranking bias behavior unchanged.

This makes Discover consistently respect the same region used for streaming availability and release windows.

### DIFF
--- a/src/lib/feed/locale.ts
+++ b/src/lib/feed/locale.ts
@@ -98,19 +98,11 @@ export function preferredLangCodes(settings: Settings): string[] {
   return [...out];
 }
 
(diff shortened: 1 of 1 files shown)


Aerya:
Bump.
@harborstremio ?

PR #1262, opened 2026-08-21
fix(localization): keep localized search results and posters from falling back to English

Thunderhawkk:
## Summary

Makes Harbor honor the user's selected metadata language consistently across search results, posters, and episode metadata. Titles, overviews, episode names, and artwork now follow the configured language with an English fallback, instead of defaulting to the original-language (e.g. Japanese) or English values that previously slipped through in different flows.

## Why

Localization was inconsistent across entry points and surfaces:

- Search was requested in the metadata language, but for titles without a translation TMDB returned blank or original-language (Japanese) info, and the localized result could be overridden by the English MAL/AniList entry once the anime search resolved causing a localized→English flicker and blank info.
- Posters fell back to the original Japanese artwork when the metadata-language poster was unavailable, instead of English.
- Anime/series episode metadata picked the first available language rather than the requested one, so episode titles/overviews ignored the metadata language (and Japanese leaked through when the language was English).

This approach fits Harbor by fixing the root cause, a language-first text picker and consistent English fallback in shared metadata/artwork logic — rather than patching individual views. It applies uniformly to search, anime, and series content.

## Verification

- npx tsc -b --pretty false — clean, exit 0 (after the TypeScript changes)
- pnpm run build — success (only pre-existing chunk-size warning)
- node --test — search-display-state, search-request-guard, and poster-backdrop-expansion tests pass
- Manual scenarios: search for a title without a localized translation (English fallback shown, no blank info); anime search keeps the localized result instead of swapping to English/MAL; posters fall back to English, not Japanese; anime/series episode names & overviews in the metadata language; English metadata language no longer shows Japanese episode text.

## Platform Impact

None

## Checklist

### DIFF
--- a/src/components/pick-card.tsx
+++ b/src/components/pick-card.tsx
@@ -240,7 +240,7 @@ const PosterCard = memo(function PosterCard({
 
   const [imgIdx, setImgIdx] = useState(0);
   const [hydratedPoster, setHydratedPoster] = useState<string | undefined>();
-  const localizedPoster = useLocalizedPoster(meta.id);
+  const { url: localizedPoster, localizing: posterLocalizing } = useLocalizedPoster(meta.id);
   const wantTmdbPoster = needsTmdbForPoster(settings.rpdbKey, meta.id);
   const resolvedTmdb = useTmdbIdFromImdb(wantTmdbPoster ? meta.id : undefined);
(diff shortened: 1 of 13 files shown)


Talal1011:
added to PR #1280

PR #1270, opened 2026-08-22
Feature/calendar release time larger posters

uddx7:
Fixes three long-standing gaps in the Calendar view:

Local release time & countdown — episode/movie air times are now converted to the user's local timezone and shown alongside the date across every calendar source (Trakt, Cinemeta, TMDB, AniList, anime dub feed, AniZip). A live "in Xd Yh Zm" countdown renders for any item that hasn't aired yet, in the month grid, day modal, and hover tooltip alike.
Larger poster option — added a Default/Large poster-size toggle in the Calendar header; grows the hover tooltip and day-modal artwork.
Blurred/placeholder posters fixed, three separate root causes:
Poster wasn't re-measuring its container on resize, so it sometimes fetched an image sized for the wrong box (ResizeObserver added).
Several calendar sources (fetchTraktCalendar, resolveSavedCalendar for both anime and non-anime library items) were prioritizing a small landscape episode still over the show's actual portrait poster, so it got stretched/cropped and looked blurry. Now the real series/movie poster is always preferred.
Anime posters from the AniZip source (My Library tab) had no series-level poster field at all — now cross-referenced against TMDB via AniZip's themoviedb_id mapping.
Also fixed a native crash on app shutdown: a diagnostic eprintln! in release_stremio_scheme would panic if stderr's pipe was already closed during teardown, taking the whole process down with it.

### DIFF
--- a/src-tauri/src/lib.rs
+++ b/src-tauri/src/lib.rs
@@ -57,11 +57,13 @@ mod web_server;
 mod webview_helpers;
 
 pub(crate) fn release_stremio_scheme(app: &tauri::AppHandle) {
+    use std::io::Write;
     use tauri_plugin_deep_link::DeepLinkExt;
-    match app.deep_link().unregister("stremio") {
-        Ok(()) => eprintln!("[harbor::deeplink] released stremio:// on shutdown"),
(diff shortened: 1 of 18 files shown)


Talal1011:
merged in #1354

PR #1305, opened 2026-08-28
Fix downloaded streams not showing as cached

uddx7:
Summary
Fixes the Downloads-tab player so a fully downloaded local file is immediately shown as Cached (green dot, no buffer-progress bar) instead of behaving like it's still streaming/buffering.

Why
The playback-downloaded effect in src/views/player.tsx only special-cased the P2P torrent engine. Every other non-live, non-HLS source — including an already-downloaded local file played from the Downloads tab — fell into a generic branch that derived "downloaded" from (position + buffered) / duration. That heuristic only crosses the 99.9% "fully cached" threshold near the very end of playback, so a file that already exists in full on disk kept showing a growing seekbar buffer and never got the green Cached dot, even though nothing was actually being downloaded.

Fix: added an isLocalSrc check (via the existing isLocalUrl helper) ahead of the generic branch — when the source is a local file, downloaded is set to 1 immediately.

Verification
pnpm run typecheck — clean
pnpm run test (tests/player-buffer-policy.test.ts) — 5/5 passing, including a new regression test asserting the local-source branch exists, runs before the generic buffer heuristic, and sets downloaded = 1
Not manually verified in a running app session (no browser/app UI pass performed)
Platform Impact
None — change is in shared TypeScript player logic (src/views/player.tsx), no native/platform-specific code touched.

UI Changes
Seekbar/time-display "Cached" green dot now appears immediately for local downloaded files, and the buffer-progress fill stays hidden. No screenshots captured.

Checklist
 Focused change, no unrelated refactors
 pnpm run typecheck (vp check/vp run typecheck scripts aren't present in this checkout)
 Cargo checks — N/A, no Rust changes
 Tested on Windows/macOS/Linux/TV input — not performed

### DIFF
--- a/src/views/player.tsx
+++ b/src/views/player.tsx
@@ -154,6 +154,7 @@ export function PlayerView({ src }: { src: PlayerSrc }) {
     (isBundledEngineUrl(src.url) || isLocalEngineUrl(src.url)) &&
     !src.url.includes("/hlsv2/") &&
     !!src.streamRef?.infoHash;
+  const isLocalSrc = isLocalUrl(src.url);
   const { stats: engineStats, genuineFailure } = useEngineStats({
     url: src.url,
     infoHash: src.streamRef?.infoHash ?? null,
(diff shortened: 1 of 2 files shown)


Talal1011:
merged in #1354

PR #1335, opened 2026-08-29
fix(simkl): reliably mark watched through end-of-playback and source switch

Thunderhawkk:
## Summary

Fixes the Simkl scrobble path so a finished movie or episode is reliably marked watched on Simkl. When playback ends (snap.status === "ended") or the source switches to the next item (key change), the terminal stop scrobble was being sent through the throttled, non-keepalive request queue. If the player closed (~800 ms after end) or the app quit, that request was dropped before it was sent. Simkl never saw the completion, so the item stayed in its "watching" state — showing back up in the Continue Watching row with a Simkl badge, stopped at whatever mid-movie progress was last reported. Fixes both terminal paths in scrobble-hook.ts.

## Why

The terminal stop is what tells Simkl to move an item out of "currently watching." Two independent gaps caused it to be lost:

1. Non-keepalive dispatch — simklScrobble("stop", ...) routes through simklRequest, which serializes on queueTail and sleeps up to 1050 ms on the POST throttle, with no keepalive. When the player unmounts (auto-end-exit closes it ~800 ms after EOF) or the app quits, that queued request can be killed in-flight.
2. No recovery — Simkl's scrobble swallows all errors, with no fallback, so a single dropped request leaves the item permanently stuck in "watching."

The fix sends the terminal stop through the teardown-safe beacon (native fetch with keepalive: true, bypassing the throttle queue) and additionally records the completion directly in Simkl's watch history as a fallback. This mirrors the keepalive handling already applied to the unmount/page-hide path (commit 35788baa) and the Trakt integration's pushWatched fallback.

## Verification

- pnpm test — 557 pass; the two new regression tests pass (completing a movie keeps the terminal stop in the teardown-safe beacon path, a watched title switched-to-next uses the teardown-safe beacon stop). 8 remaining failures are a pre-existing baseline unrelated to this change.
- pnpm run typecheck — 0 TypeScript errors.
- Confirmed before the fix, the natural-end regression test failed (natural-end stop must use the keepalive beacon) and passed after.

## Platform Impact

None

## Checklist

### DIFF
--- a/src/lib/simkl/scrobble-hook.ts
+++ b/src/lib/simkl/scrobble-hook.ts
@@ -1,6 +1,8 @@
 import { useEffect, useRef } from "react";
 import { useSimkl } from "./provider";
 import { simklScrobble, buildBody, type ScrobbleInfo } from "./scrobble";
+import { addToHistory, markEpisodesWatched } from "./history";
+import { stremioIdToSimklTarget } from "./ids";
 import { getPlaybackPosition } from "@/lib/player/playback-clock";
 import { useSettings } from "@/lib/settings";
 import type { PlayerSrc } from "@/lib/view";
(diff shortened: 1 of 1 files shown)


Talal1011:
merged in #1354

PR #1339, opened 2026-08-30
fix(cw): honor metadata addon for CW artwork and keep rewatched movies when signed out out stremio

Thunderhawkk:
## Summary

1. CW artwork now respects your metadata addon. ContinueCard previously resolved its image through the raw Cinemeta call, which short-circuits to null whenever "Use Cinemeta for title metadata" is toggled off. It now routes through resolveMeta (src/lib/meta-resource.ts), which prefers an installed custom metadata addon, respects "Prefer my installed metadata addon," and falls back to (forced) Cinemeta — so the card no longer degrades to a stretched poster/backdrop.

2. Rewatching a finished movie returns it to Continue Watching without Stremio sign-in. The watched-flag reset lived only in the Stremio cloud write (gated on authKey). When not signed in, flaggedWatched never cleared, so a rewatched movie stayed hidden. The reset is now done locally on a meaningful rewatch (≥45s, below the finish ratio) via setMovieWatchedLocal + a local CW entry.

## Why

continue-card.tsx called meta(type, id) without force, so cinemetaEnabled=false made it return null immediately (src/lib/cinemeta.ts:132). That left metaBg/logo/hydratedMeta unset and the card fell back to item.background (backdrop) or a 2:3 poster crammed into a 16:9 box. Routing through resolveMeta honors the addon preference (the intended reason a user toggles Cinemeta off) and still falls back to Cinemeta when no addon has data.

use-resume-autosave.ts previously wrote CW progress only for non-cloud ids. When Harbor isn't signed into Stremio, the cloud write that normally resets flaggedWatched on a rewatch never runs, so a rewatched finished movie was filtered out by isCwMember. Treating the rewatch locally (pure localStorage) fixes it independent of authKey, gated on isMovieWatchedLocal so it doesn't broaden into "all cloud movies appear when signed out."

## Verification

- pnpm run typecheck (tsc -b --pretty false) — exit 0, no errors
- Manual scenario (recommended, not executable in this headless env): with Harbor signed out of Stremio, rewatch a previously-finished movie past ~45s but below 85% → it should reappear in Continue Watching; mark/unmark watched → item leaves/returns with progress intact.

## Platform Impact

None

## Checklist

### DIFF
--- a/src/components/continue-card.tsx
+++ b/src/components/continue-card.tsx
@@ -2,7 +2,8 @@ import { memo, useEffect, useMemo, useRef, useState } from "react";
 import { Check, X } from "lucide-react";
 import { Play } from "@/components/icons/play-filled";
 import simklLogo from "@/assets/simkl.png";
-import { meta as fetchMeta, narrowMediaType, type Meta } from "@/lib/cinemeta";
+import { narrowMediaType, type Meta } from "@/lib/cinemeta";
+import { resolveMeta } from "@/lib/meta-resource";
 import { animeKitsuMeta, type AnimeKitsuVideo } from "@/lib/providers/anime-kitsu-addon";
 import { tmdbLiteMeta } from "@/lib/providers/tmdb/tmdb-lite";
 import { tmdbIdFromImdb } from "@/lib/providers/tmdb";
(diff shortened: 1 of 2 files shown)


harborstremio:
merged in #1354

PR #1343, opened 2026-08-31
feat: clickable theme-creator name → profile, with hover preview

uddx7:
## Theme Creator Profile Linking

This PR completes the missing half of the `theme-creator-handle-sync` fix.

Previously, updating a theme creator's `@handle` could leave the creator information displayed on themes outdated or disconnected from their current profile. With the backend now providing a live `authorHandle`, this change makes the creator's name clickable and hoverable across all theme surfaces, allowing users to open the creator's profile.

The implementation reuses the existing `UserHoverCard` pattern from `market-card.tsx` rather than introducing a new implementation.

### Changed Files

* `theme-detail.tsx` — Full theme details modal, which is the screen referenced in the original bug report.
* `community-detail.tsx` — Legacy theme details modal.
* `community-browser.tsx` — Theme grid cards.
* `community-store/market/market-hero.tsx` — Featured theme hero card.
* `community-store/store-top-charts.tsx` — Trending, Popular, and New theme rows.

### Implementation

Where `authorHandle` is available, the creator name is wrapped with:

`<UserHoverCard handle={authorHandle}>`

This makes the creator name interactive and allows users to open their profile.

If `authorHandle` is not available, the UI falls back to plain text. This covers cases such as anonymous or unclaimed uploads and follows the same conditional pattern already used in `market-card.tsx`.

In `market-hero.tsx`, the existing `"swatch" in item` type guard is used because `authorHandle` exists on `StoreTheme` but not on `StoreBundle`.

### Additional Regressions Fixed

During live testing, two additional interaction issues were identified and fixed:

1. **Card click propagation**

   Clicking the profile interaction inside a theme card also triggered the card's `onClick`, causing the theme details page to open behind the profile.

   Fixed by adding `e.stopPropagation()` in:

   * `community-browser.tsx`
   * `store-top-charts.tsx`

2. **Theme modal remaining open**

   Opening a creator profile from inside `ThemeDetail` or `CommunityDetail` left the theme details modal open over the profile page.

   Both modals now subscribe to `subscribeOpenProfile` and call `close()` when a profile is opened, following the same pattern already used in `add-friends-modal.tsx`.

### Backend Dependency

This change depends on the `harbor-hosted` commit:

`6814fcb` — `feat(harbor-themes): enhance theme ownership details in public view`

The backend now returns a live `authorHandle` and `author` resolved from the creator's current account instead of relying on the frozen string stored when the theme was originally uploaded.

Without this backend change, `authorHandle` will always be `null`, causing all theme surfaces to fall back to plain, non-clickable creator text.
befor
لقطة شاشة 2026-08-31 142018

### DIFF
--- a/src/views/settings/theme-panel/custom-themes-section/community-browser.tsx
+++ b/src/views/settings/theme-panel/custom-themes-section/community-browser.tsx
@@ -2,6 +2,7 @@ import { useEffect, useState } from "react";
 import { AlertCircle, Check, Download, Loader2, Star, Upload } from "lucide-react";
 import { Search } from "@/components/icons/search-icon";
 import { browseThemes, downloadTheme, rateTheme, type StoreTheme } from "@/lib/theme-store";
+import { UserHoverCard } from "@/views/profile/user-hover-card";
 import { CommunityDetail } from "./community-detail";
 import { ThemeUploadFlow } from "./theme-upload-flow";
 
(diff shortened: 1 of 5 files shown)


Talal1011:
merged in #1354

PR #1345, opened 2026-08-31
fix(addons): keep addon position and avoid duplicates on update

Thunderhawkk:
## Summary

When an addon was updated, Harbor pushed a duplicate copy to the end of the collection instead of replacing the original, which also dropped the addon to the bottom of the ordering — both in the local Installed tab and in the Stremio account collection.

The update flow now resolves the addon being replaced by its manifest id, transport URL, or explicit replaceId, removes only those entries, and re-inserts the new addon at the original's position — matching Stremio's own update behavior. Local display order is preserved through harbor.addonOrder via a new preserveOrderOnReplace helper.

## Why

- Duplicates: installAddon/installFromUrl only deduplicated by transport URL, so an update with a changed URL (or a matched replaceId handled after the fact) produced a second entry.
- Position reset: pushToStremio blindly appended the updated addon to the end of the cloud collection. For addons synced from a Stremio account — which are read from the cloud and never present in loadInstalled() — the replacement slot was always empty, so the addon fell to the bottom even though Stremio itself preserves position on direct updates. Keeping the position on reinstall is the expected, predictable behavior for the app.

The dedup/splice logic lives in pushToStremio (against the cloud collection) and a shared preserveOrderOnReplace helper (against the local order), keeping the ordering concern in the addon-store layer rather than scattered across views.

## Verification

- pnpm run typecheck (tsc -b --pretty false) — exit 0, clean.
- Reviewed the splice logic for index alignment when the replacement removes multiple entries before the insert index.
- Confirmed the position is now resolved from the Stremio collection by manifest.id, so cloud-synced addons (not present in loadInstalled()) keep their position.
- Manual scenario to confirm: sign in to Stremio in Harbor, update an addon, and verify no duplicate appears and the addon stays in the same place in the Installed tab and the Organize/reorder page.

## Platform Impact

None

## Checklist

### DIFF
--- a/src/lib/addon-store.ts
+++ b/src/lib/addon-store.ts
@@ -1,7 +1,11 @@
 import { safeFetch as fetch } from "@/lib/safe-fetch";
 import { readActiveStremioAuthKey } from "./auth";
 import { setUserAddons, userAddons, type Addon } from "./addons";
-import { applyOrderToItems } from "./addons-store/reorder";
+import {
+  applyOrderToItems,
+  loadDisplayOrder,
+  saveDisplayOrder,
+} from "./addons-store/reorder";
 
 const PROFILES_KEY = "harbor.profiles.v1";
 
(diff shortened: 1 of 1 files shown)


Talal1011:
merged in #1354

PR #1346, opened 2026-08-31
fix: unresponsive hero/context-menu buttons

uddx7:
## Summary

* Fix the Home hero banner's action buttons (Add to Watchlist, Play, More info,
  nav arrows) going permanently unresponsive after the first drag/swipe on the
  hero carousel — `HeroCarousel`'s `moved` ref was never reset after a drag
  ended, so `onClickCapture` kept swallowing every subsequent click on any
  button inside it.
* Fix the "Add to my list" flyout in the right-click context menu never
  appearing — it was positioned `absolute` outside the bounds of an ancestor
  with `overflow-y-auto`, which CSS silently clips on both axes once either
  axis is non-`visible`. Now renders through a `document.body` portal with
  viewport-computed coordinates, matching the same pattern already used by
  `AddToListMenu`/`AnchoredMenu` elsewhere in the app.

## Test plan

* [x] Drag/swipe the Home hero banner, then click Play / More info / Add to
  Watchlist / the nav arrows — all should respond immediately.
* [x] Right-click a title while it's playing, hover/click "Add to my list" —
  the flyout list of custom lists should appear beside it.
* [x] `pnpm run typecheck` and `cargo check --manifest-path src-tauri/Cargo.toml`
  both pass clean (verified this session).

### DIFF
--- a/src/components/context-menu/my-list-submenu.tsx
+++ b/src/components/context-menu/my-list-submenu.tsx
@@ -1,5 +1,6 @@
 import { Check, ChevronRight, ListPlus, Plus } from "lucide-react";
 import { useRef, useState } from "react";
+import { createPortal } from "react-dom";
 import {
   addToList,
   toggleInList,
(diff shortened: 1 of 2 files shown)


Talal1011:
merged in #1354

PR #1347, opened 2026-08-31
feat(profile-import): add merge/replace conflict choices when importing from primary

Thunderhawkk:
## Summary

Adds per-domain conflict handling to the existing "Import data from primary profile" flow. When a selected area (watchlist, favorites, or addons) already contains items on the target profile, the import panel now reveals a Merge / Replace toggle for that domain instead of silently overwriting. Merge is the default and never loses existing data; Replace fully overwrites the domain with the primary's version.

## Why

Previously, importing from the primary unconditionally replaced each selected area on the target profile, which could wipe data the user had already added to the target (e.g. a watchlist entry present only on the target). The user asked for merge semantics so that importing brings in what the primary has and keeps what the target already has.

## Verification

- pnpm run typecheck passes cleanly (tsc -b --pretty false).
- Manually traced merge (union by id/transportUrl, target-wins) and replace (wholesale overwrite) paths in importDomains(), including the addons disabled-list union and the selected-subset addon merge fix.
- Fix verified: the merge path now imports only the user-selected addon subset rather than all source addons

## Platform Impact

None

## UI Changes

A Merge / Replace toggle row appears per mergeable domain only when that domain has overlapping items on the target profile (defaulting to Merge). When there are no overlaps, no extra UI is shown.

image

## Checklist

### DIFF
--- a/src/components/profile-picker/editor-view.tsx
+++ b/src/components/profile-picker/editor-view.tsx
@@ -39,11 +39,14 @@ import {
 } from "@/lib/profiles";
 import { emitListToast } from "@/components/lists/list-toast";
 import {
+  analyzeOverlaps,
   defaultSelectedAddonUrls,
   importDomains,
   summarizeSource,
+  type DomainOverlap,
   type ImportAddonPreview,
   type ImportDomain,
+  type ImportDomainChoice,
   type ImportSourceSummary,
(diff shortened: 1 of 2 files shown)


Talal1011:
merged in #1354

PR #1349, opened 2026-08-31
fix and feat: create theme filtering and improve friend presence

pengunnn:
## Summary

- Add behavior-based filtering to the community theme store.
- Keep the filter menu within the available viewport and make overflowing options scrollable.
- ensuring each friend’s status and indicator match between the friends panel and hover card.

## Why

The community theme store needed a way to filter themes by navigation behavior. The filter menu also became partially inaccessible in smaller windows when it opened outside the available viewport.

Friend rows and hover cards could independently resolve presence information, causing them to display different statuses. Passing the normalized friend presence into the hover card keeps both views synchronized

## Verification

- Ran `git diff --check`.
- Verified the filter opens above or below the button according to available space.
- Verified the filter menu remains scrollable in a small window.
- Verified behavior filters update the displayed theme results.
- Verified friend-row and hover-card presence dots and labels match.

### DIFF
--- a/src/lib/social/presence.ts
+++ b/src/lib/social/presence.ts
@@ -68,6 +68,14 @@ async function savePresence(next: PresenceStatus): Promise<void> {
   });
 }
 
+export function resolvePresence(value: unknown, online = false): PresenceStatus {
+  if (value === "online" || value === "away" || value === "dnd" || value === "offline") {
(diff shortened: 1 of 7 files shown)


Talal1011:
merged in #1354

PR #1353, opened 2026-09-01
fix: seed initial player volume from saved prefs to avoid stale baseline

Thunderhawkk:
## Summary

Every fresh player bridge initialized its volume state to a hard-coded 100% (volume: 1) via emptySnapshot, and only corrected it once mpv emitted a volume property-change event. When mpv was already at the persisted volume, the restore write was a silent no-op (libmpv emits no property-change for same-value sets), so the event never arrived and the frontend baseline stayed at 100% — making the first scroll-wheel or keyboard volume step jump to ~95% instead of stepping down from the saved value.

This change seeds the initial snapshot on every fresh player session from the persisted prefs (readPlayerVolume()), so a no-op restore can no longer leave a stale 100% baseline. Volume persistence itself is unchanged

## Why

- Bug (log-confirmed): with a retained mpv instance already at the saved volume, the restore effect's setVolume(savedVolume) was a same-value no-op; libmpv sent no property-change event, so snap.volume stayed at the emptySnapshot default of 1.0. The first wheel/keyboard step then computed 1.0 − 0.05 = 95%, matching the reported "at 50%, going down goes to 90%" symptom. Verified in the mpv log: restore volume="60" followed by first wheel volume="95".
- Why this approach fits Harbor: it treats the restore value as the truthful initial state (the restore always pins mpv to the saved value), keeps mpv as the source of truth for changes, and requires no new events or protocol changes. It aligns with the existing optimistic-update pattern already used by the wheel

## Verification

- pnpm run typecheck (tsc -b) — clean on all changed files.
- Manually confirmed fix behavior across scenarios: new user (nothing saved → defaults to 100%, unchanged); saved 0.5 → bridge starts at 0.5; boosted volume (2.0) preserved; corrupt saved JSON → graceful fallback to 100%; bridge reuse across episodes unchanged.
- Reproduced the no-op-restore path in code review: retained-mpv-already-at-saved-value now leaves the frontend at the correct baseline instead of 100%

## Platform Impact

None

## Checklist

### DIFF
--- a/src/lib/player/bridge.ts
+++ b/src/lib/player/bridge.ts
@@ -1,3 +1,4 @@
+import { readPlayerVolume } from "@/lib/player-volume";
 import type { SubCue } from "@/lib/subtitles/parser";
 import type { SubtitleLoadMetadata } from "@/lib/subtitles/types";
 import type { SubtitleMatchConfidence } from "@/lib/subtitles/release-match";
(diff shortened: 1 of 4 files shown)


harborstremio:
merged in #1354

PR #1374, opened 2026-09-03
fix: prevent series from auto-adding on page load

Thunderhawkk:
## Summary

Stops a series from being auto-added to the watchlist the moment its detail page loads. The change in src/views/detail.tsx initializes prevSeriesWatchedVerRef to the current watched version instead of a stale -1, so the page-load effect no longer fires on the first render and no longer triggers the add.

## Why

Opening any series detail page added the series to the watchlist ("watch later") immediately, even though the user never clicked Add to Watchlist. The -1 sentinel made the first comparison always differ from the current version, so the effect ran on mount. Initializing the ref to the live value skips that spurious first-run add while preserving the correct behavior when the user actually interacts with the watched/watchlist state afterwards

## Verification

- Reproduced on a local build where a series was auto-added to the watchlist the moment its detail page opened.
- After the change, opening the same series detail page no longer adds it to the watchlist.
- Confirmed the add still happens when the user explicitly clicks Add to Watchlist.
- vp check passes for src/views/detail.tsx.
- vp run typecheck passes.

## Platform Impact

None

## Checklist

### DIFF
--- a/src/views/detail.tsx
+++ b/src/views/detail.tsx
@@ -1108,7 +1108,7 @@ export function DetailView({
     manualWatchedVersion,
     manualWatchedVersion,
   );
-  const prevSeriesWatchedVerRef = useRef(-1);
+  const prevSeriesWatchedVerRef = useRef(seriesWatchedVer);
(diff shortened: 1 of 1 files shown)


Talal1011:

PR #1382, opened 2026-09-04
fix(subtitles): preserve the latest manual selection

Talal1011:
## Summary
- keep the latest manual subtitle selection authoritative while provider and autoload results arrive
- serialize native mpv subtitle transitions and confirm the selected track from mpv before updating UI state
- prevent delayed automatic, restore, or cache work from replacing a newer manual selection or manual Off
- restore only an exact remembered subtitle identity and normalize cached Windows paths
- publish immutable player snapshots so consecutive cues, clears, and seeks render without unrelated input

## Reported regression covered
Selecting subtitle 5, then subtitle 6, and leaving the panel while results are still arriving must retain subtitle 6. An older cache completion can no longer restore subtitle 5.

## Validation
- 72 focused subtitle tests: 71 pass, one optional native fixture test skipped
- TypeScript project check passes
- changed-file formatting/lint passes

## Rollout hold
This is intentionally a draft. Do not merge it into the public beta until this commit passes as tester-facing Experimental 0.0.1, using internal experimental package 0.999.1 with tested return to beta 0.9.124.


Talal1011:

PR #1394, opened 2026-09-06
fix(fullscreen): restore window bounds and maximized state on exit

Thunderhawkk:
## Summary

Fixes window state being clobbered in two lifecycle stages: first launch and fullscreen/borderless transitions.

Commit 1 — fit_to_monitor no longer overwrites saved state (src-tauri/src/display_fit.rs):

- Skips fitting when the window is already maximized (sizing calls would un-maximize it).
- Skips fitting on launch when the window-state plugin has a non-default saved geometry — the plugin already restored it during window creation, so fitting would clobber it.
- Adds a Windows work-area inset helper so fitting accounts for the taskbar (logical insets, matching macos_screen_insets behavior).

Commit 2 — fullscreen/borderless exit restores the saved window (src-tauri/src/fullscreen.rs, src/lib/fullscreen-state.ts):

- window_fullscreen_enter: captures the maximized flag before the async unmaximize(); previously reading it after still reported true and dropped the saved geometry for maximized windows.
- window_fullscreen_exit: restores state even when the window was already unfullscreened (native Esc / exitAnyFullscreen paths bypass the if is_fs guard). Re-maximizes when it was maximized, otherwise restores size/position or centers per fullscreenRestorePosition.
- enterBorderless/exitBorderless: same maximized race fixed on the borderless path; exit restores via setMaximized(true) (granted toggleMaximize) instead of win.maximize(), which Tauri capabilities would reject at runtime.

## Why

Two distinct bugs left the window where the user didn't put it:

On launch with a saved window state, fit_to_monitor overwrote the restored geometry (and un-maximized a maximized window).
Leaving fullscreen/borderless could restore to fullscreen/maximized bounds — the saved geometry was dropped for maximized windows, and the Esc path skipped restore entirely, leaving the window off-screen at negative coordinates.
Together the change keeps the window where the user left it: restored state survives launch, and every fullscreen exit path lands back on the prior bounds/state.

## Verification

- cargo check --manifest-path src-tauri/Cargo.toml — PASS (2 pre-existing vapoursynth warnings)
- cargo test --manifest-path src-tauri/Cargo.toml --lib — 174 passed, 0 failed
- pnpm run typecheck (tsc -b --force --pretty false) — PASS
- pnpm run test — 985/1014 pass
- player-fullscreen-modes.test.ts — 7/8 pass, matches HEAD baseline

## Platform Impact

- Windows: new work-area inset path for fitting + the async unmaximize restore fix (most visible). Needs manual smoke test.
- macOS: macos_screen_insets path unchanged; fullscreen exit restore is shared code — smoke test recommended.
- Linux: fitting unchanged (0.0 insets as before); fullscreen restore is shared code.

## Checklist


Talal1011:

PR #1422, opened 2026-09-13
fix(player): load mpv scripts before init; place separate-window player on Harbor's display

Thunderhawkk:
## Summary

Fixes Windows HDR script loading and separate-window display placement:

Script loading: scripts, script-opts, and load-scripts are init-only mpv options that cannot be set after mpv_initialize. extra_options are now split into init-only and runtime pairs so scripts (e.g. hdr-mode.lua, display-info.lua) are applied before initialization.

Separate-window placement: the separate-window player always opened on the primary monitor. It now opens on the monitor where the Harbor main window sits, by resolving the main window's monitor index and setting mpv's screen option before init.
Removed dead vo-window tracking: drops MPV_VO_MONITOR / capture_mpv_vo_monitor — the vo window is destroyed before restore runs, so the feature could never work. SDR restore now simply targets the Harbor main window's monitor.

Known limitation (intentional, not addressed here): separate-window mode still does not flip the display back to SDR on exit. That behavior is left as-is per the current scope; only embedded-mode exit restore is guaranteed.

## Why

- The separate-window mpv window ignored which display the user was actually using, landing on the primary monitor instead of the TV.
- The vo-window monitor tracking for HDR restore was fundamentally broken (referenced a window destroyed before restore) and was removed rather than papered over.
- apply_pre_init previously couldn't express script configuration at all because script options are rejected after init; splitting extra_options gives the pre-init pass the missing capability.
- This fits Harbor's architecture: Rust owns OS/display behavior, and the frontend only re-syncs its HDR stage state after the OS-level flip lands.

## Verification

- cargo check --manifest-path src-tauri/Cargo.toml — clean
- cargo test --manifest-path src-tauri/Cargo.toml --lib extra_option_parser — 8 passed.
- cargo test --manifest-path src-tauri/Cargo.toml --lib script_pre_init — 5 passed.
- pnpm run typecheck
- Manual (pending TV retest): separate-window player should open on the monitor showing the Harbor main window. With HDR mode lua script Embedded-mode HDR-off-on-exit restore behavior is unchanged.

## Platform Impact

- Windows: primary target — EnumDisplayMonitors index resolution, pre-init screen set, DisplayConfig-based HDR flip/restore (Win32_Devices_Display feature added).

## Checklist

- [x] I tested affected platforms when the change is platform-specific.

### DIFF
--- a/src-tauri/Cargo.toml
+++ b/src-tauri/Cargo.toml
@@ -105,6 +105,7 @@ windows = { version = "0.61", features = [
   "Win32_Graphics_Gdi",
   "Win32_Graphics_Dxgi",
   "Win32_Graphics_Dxgi_Common",
+  "Win32_Devices_Display",
   "Win32_System_Threading",
   "Win32_System_Power",
   "Win32_System_JobObjects",
--- a/src-tauri/src/mpv.rs
+++ b/src-tauri/src/mpv.rs
@@ -155,6 +155,10 @@ struct MpvSession {
     is_live: bool,
     #[cfg(any(windows, target_os = "linux"))]
     embedded: bool,
+    /// OS-level HDR state when the session started. Compared against the
+    /// script-reported state at teardown: only an off->on flip waits.
+    #[cfg(windows)]
+    hdr_baseline_on: bool,
 }
 
 impl MpvState {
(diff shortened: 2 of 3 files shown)


Talal1011:

PR #1442, opened 2026-09-20
fix(linux): create the mpv render context before loadfile

kalashnikxvxiii:
## Summary

On Linux, opening a title that loads quickly (anything cached, or served by
the local proxy) plays audio only: mpv logs `No render context set` and drops
the video track for the whole session. This creates the render context before
`loadfile` runs, so the race can't happen.

## Why

It's a race between two things Harbor starts at the same time. `install()`
built the render context lazily, inside the GLArea's render callback, so it only
existed once GTK had painted the area for the first time. But `mpv_start` does
not wait for that: it issues `loadfile` the moment `install()` returns. If the
file opens before that first paint, mpv brings up the VO with no render
context and gives up on video.

Measured on one playback, from the mpv log:

| time | event |
|---|---|
| 177.773 s | render context creation starts |
| 177.810 s | mpv opens the file, looks for the context, finds none |
| 178.094 s | context finishes building, too late |

The window is about 320 ms, and around 240 of those go to probing the VA-API
decoder. Which is why it looked random: a stream that takes seconds to open
always won the race, a cached file that opens in milliseconds always lost it.
Heavy shaders made it worse from the other side, by stretching how long the
context takes to build.

The fix builds the context inside `install()`, right after `show_all()` has
realized the area and its GL context can be made current. That's what the
macOS path already does. The lazy path in the render callback stays as a
fallback: if eager creation ever fails, we're back to today's behaviour, with a
log line saying so.

## Verification

- `cargo check` clean, no new warnings (16 before and after, none in this file)
- `cargo test` 202 passed, 0 failed, 2 ignored
- Source build, two playbacks of a locally proxied title, the case that lost the
  race every time: video shows, and the log has zero `No render context set`,
  zero `Video: no video`, and no fallback messages. `render surface` is logged
  before `loadfile`, so the context was there when mpv opened the file

## Platform Impact

Linux only, verified on Wayland with NVIDIA. Windows and macOS are untouched:
the change is entirely inside `mpv_render_linux.rs`.

## Checklist

- [x] I tested affected platforms when the change is platform-specific. (Linux)
- [ ] I added or updated tests for behavior changes. (needs a live GTK window and GL context; verified manually as above)


Talal1011:
Reviewed c8d1eea877d5c01e69f516084b2bebfb1bc75ede against beta. The PR remains a 24-line Linux-only startup change: initialize the render context before install returns/loadfile, retaining the lazy fallback. Linux CI cargo check passed; Windows cargo check passed with existing dead-code warnings. Linux native tests: 603 passed, 2 failed, 2 ignored. The two failures are existing music fixture teardown errors, not this GTK startup path: native_meter_reads_real_stereo_preserves_processing_and_disables and native_seek_to_end_can_resume_and_paused_seeks_keep_their_position retain their local Arc<Mpv> while calling engine.stop(false), but retire_session requires Arc::try_unwrap before its 5-second deadline. The fixtures only drop that reference after stop returns. The entire music/engine.rs blob is identical to the beta base, and these null-output fixtures do not call mpv_render_linux::install. Frontend format/lint/typecheck passed; its two baseline test failures remain translation coverage and the obsolete Sports restriction assertion. Production frontend build passed. Full Linux packaging was attempted locally but blocked by the absent Linux toolchain; no live Linux display/GPU test was possible here. Keeping unrelated baseline repairs out of this PR; not claiming a fully green suite.

